PR #190, opened 2021-02-24
ChangeValidator Era Notices, Require Set Keys

maxwolff:
admin, era changing notices such as (set_yield, set_supply_cap, change_authority)  now which bump the notice era and set a `NoticeHold` on any new notices being signed until they are signed. The notices are added to`PendingEraNotices`, if they are executed they are removed upon `on_initialize`.

* adds `shouldEndSession`, which is periodic unless an era change is ocurring
* tests the `author_rotateKeys` => `session_setKeys` => `changeAuthorities` flow, and make sure changeAuthorities fails if account does not have grandpa/aura keys set
* goes with this 1 line substrate change https://github.com/compound-finance/substrate/pull/27


integration/__tests__/gov_test.js
```diff
@@ -31,8 +32,8 @@ buildScenarios('Gov Scenarios', gov_scen_info, [
     scenario: async ({ ctx, chain, starport, validators }) => {
       const alice = validators.validatorInfoMap.alice;
       const alice_account_id = alice.aura_key;
-      const newAuthsRaw = [[ctx.actors.keyring.decodeAddress(alice_account_id), { eth_address: alice.eth_account }]];
-      let extrinsic = ctx.api().tx.cash.changeAuthorities(newAuthsRaw);
+      const newAuthsRaw = [{ substrate_id: ctx.actors.keyring.decodeAddress(alice_account_id), eth_address: alice.eth_account }];
```
maxwolff: format changes in a prev pr broke this test

pallets/cash/src/core.rs
```diff
@@ -335,15 +335,31 @@ pub fn apply_chain_event_internal<T: Config>(event: ChainLogEvent) -> Result<(),
...
+    should_increment_era: bool,
     notice_fn: &dyn Fn(NoticeId, ChainHash) -> Result<Notice, Reason>,
 ) -> Result<(), Reason> {
     // XXX this cannot fail, should not return result
     let (latest_notice_id, parent_hash) =
         LatestNotice::get(chain_id).unwrap_or((NoticeId(0, 0), chain_id.zero_hash()));
-    let notice_id = latest_notice_id.seq();
+
+    let notice_id = if should_increment_era {
+        require!(
```
jflatow: So we return an error on the extrinsic if there are any pending era notices to be signed? Need to think about this, didn't we ideally want to continue accepting transactions, just not signing them yet?

jflatow (commented):
Great stuff

pallets/cash/src/params.rs
```diff
@@ -19,3 +19,10 @@ pub const TRANSFER_FEE: Quantity = Quantity::from_nominal("0.01", CASH);
 
 /// Number of blocks between HTTP requests from offchain workers to open oracle price feed.
 pub const ORACLE_POLL_INTERVAL_BLOCKS: u32 = 10;
+
+/// The number of blocks to wait before switching sessions
+#[cfg(debug_assertions)]
+pub const SESSION_PERIOD: u32 = 2;
```
maxwolff, line 25: just noticed we set this to 50 before, makes integration tests go slower, what was the reason behindthis?

coburncoburn: can be 2 in test, but running live it is really spammy to be changing sessions every 2 blocks

pallets/cash/src/lib.rs
```diff
@@ -329,6 +359,53 @@ impl<T: Config> pallet_session::SessionManager<SubstrateId> for Module<T> {
...
+        } else {
+            // no era changes pending, periodic
+            let period: T::BlockNumber = <T>::BlockNumber::from(params::SESSION_PERIOD as u32);
+            (now % period) == <T>::BlockNumber::from(0 as u32)
+        }
+    }
+}
+
+impl<T: Config> frame_support::traits::EstimateNextSessionRotation<T::BlockNumber> for Module<T> {
+    fn estimate_next_session_rotation(now: T::BlockNumber) -> Option<T::BlockNumber> {
```
coburncoburn, line 450: is this just part of the interface, or used somehow?

maxwolff: its just part of the interface, not used for anything afaict

pallets/cash/src/core.rs
```diff
@@ -1381,6 +1399,14 @@ pub fn on_initialize_internal<T: Config>(
     LastMinerSharePrincipal::put(miner_share_principal);
     LastYieldTimestamp::put(now);
 
+    // undo the hold placed on notice signing during era changes
+    for (chain_id, notice_id) in PendingEraNotices::iter() {
```
coburncoburn: on initialize internal is getting a bit large

pallets/cash/src/internal/change_validators.rs
```diff
@@ -1,14 +1,43 @@
...
+    chains::{ChainHash, ChainId},
+    core::dispatch_notice_internal,
+    notices::{ChangeAuthorityNotice, Notice},
+    reason::Reason,
+    require,
+    types::ValidatorKeys,
+    Config, Event, Module, NextValidators, SessionInterface,
+};
 
 pub fn change_validators<T: Config>(validators: Vec<ValidatorKeys>) -> Result<(), Reason> {
```
coburncoburn, line 13: who calls change_validators?

maxwolff: called here, must be root. so either sudo or by gov. https://github.com/compound-finance/compound-chain/blob/e17c2bad5ae7532ffe72f8ea9a1334b4e8fd3eac/pallets/cash/src/lib.rs#L447

integration/util/substrate.js
```diff
@@ -74,6 +74,15 @@ function waitForEvent(api, pallet, method, opts = {}) {
   return promise;
 }
 
+// TODO
```
hayesgm: You can now use `sendAndWaitForEvents(call, api, { sender: signer })`

pallets/cash/src/internal/notices.rs
```diff
@@ -27,30 +27,37 @@ pub fn handle_notice_invoked<T: Config>(
...
-                        .ok_or(Reason::NoticeMissing(chain_id, notice_id))?;
-                    let signature: ChainSignature = notice.sign_notice()?;
-                    log!("Posting Signature for [{},{}]", notice_id.0, notice_id.1);
-                    let call = <Call<T>>::publish_signature(chain_id, notice_id, signature);
-
-                    // TODO: Do we want to short-circuit on an error here?
-                    let _res =
-                        SubmitTransaction::<T, Call<T>>::submit_unsigned_transaction(call.into())
-                            .map_err(|()| Reason::FailedToSubmitExtrinsic);
+        let mut notice_hold = false;
```
hayesgm: Why not just take `NoticeHold` and then compare the notice ids (making sure that notices implement ordering?)

pallets/cash/src/lib.rs
```diff
@@ -175,6 +174,15 @@ decl_storage! {
         /// The most recent notice emitted for a given chain
         LatestNotice get(fn latest_notice_id): map hasher(blake2_128_concat) ChainId => Option<(NoticeId, ChainHash)>;
 
+        // Notices that change the starport era. All pending era notices must be signed before new regular notices can be signed
```
hayesgm: Can't notice hold just handle this semi-magically?

pallets/cash/src/internal/notices.rs
maxwolff: not sure i understand

pallets/cash/src/core.rs
```diff
@@ -374,15 +374,26 @@ pub fn apply_chain_event_internal<T: Config>(event: ChainLogEvent) -> Result<(),
...
+    let notice_id = if should_increment_era {
+        require!(NoticeHold::iter().count() == 0, Reason::PendingEraNotice);
+        latest_notice_id.seq_era()
+    } else {
+        latest_notice_id.seq()
+    };
 
     let notice = notice_fn(notice_id, parent_hash)?; // XXX fixme cannot fail
 
+    if should_increment_era {
```
hayesgm, line 393: Is this always true? Can we have new eras without a hold? I feel like this is a separate var "should hold"?

pallets/cash/src/lib.rs
```diff
@@ -329,6 +353,111 @@ impl<T: Config> pallet_session::SessionManager<SubstrateId> for Module<T> {
...
+        a_set.insert(v);
+    }
+
+    let mut b_set = BTreeSet::<T>::new();
+    for v in b {
+        b_set.insert(v);
+    }
+
+    let count = a_set.intersection(&b_set).into_iter().count();
+    println!("a={}, b={}, count={}", a_count, b_count, count);
```
hayesgm, line 388: I think my recent patch fixes this (the `recover_user_address` versus `recover_address` fix)

pallets/cash/src/lib.rs
```diff
@@ -329,6 +353,111 @@ impl<T: Config> pallet_session::SessionManager<SubstrateId> for Module<T> {
...
+                true
+            } else {
+                println!("should_end_session=false[notices_held]");
+                false
+            }
+        } else {
+            // no era changes pending, periodic
+            let period: T::BlockNumber = <T>::BlockNumber::from(params::SESSION_PERIOD as u32);
+            let is_new_period = (now % period) == <T>::BlockNumber::from(0 as u32);
+            println!(
```
hayesgm, line 440: We might not want to print every time, but it really helps while we're testing to have this. In the future this probably should, at the least, only show when it's `true` for the periodic check.

hayesgm (approved):
Looks _excellent_, only question is around whether we should always hold on new eras.

PR #209, opened 2021-03-01
Lock To Any

hayesgm:
This patch adds Lock To Any, which allows a user on Ethereum to lock an asset to any chain account, not just an Ethereum account. For instance, you could call `lockTo(..., 'SOL', "5eff5e...");`. We interpret these values correctly downstream to pull in any given chain.


jflatow (approved):
🔥

ethereum/contracts/Starport.sol
```diff
@@ -467,4 +478,8 @@ contract Starport {
         }
         return abi.decode(_returnData, (string)); // All that remains is the revert string
     }
+
+    function toBytes32(address addr) public pure returns (bytes32) {
```
antonina-cherednichenko, line 482: Find the whole list of interesting methods for conversion https://ethereum.stackexchange.com/questions/884/how-to-convert-an-address-to-bytes-in-solidity

antonina-cherednichenko (approved):

ethereum/contracts/Starport.sol
hayesgm: Yeah, none of them really worked well for me, but this one did and it's the most obvious.

PR #212, opened 2021-03-03
Revert "Lock To Any"

jflatow:
This reverts commit 517ec985d201b66b3df41af922e40cebd0daccce.

The commit introduced a breaking change and we will use the opportunity to rehearse approaching this type of situation appropriately.


hayesgm (approved):
lgtm

PR #216, opened 2021-03-03
remove account id from types.json

coburncoburn:
polkadot.js needs to use account id to read nonces, and this is incompatible


jflatow (approved):
🔥 🚒 🧯

maxwolff (approved):

PR #210, opened 2021-03-02
Fix Lock Scenarios

hayesgm:
This patch fixes and builds out just about all of the rest of the lock scenarios. This is one of many patches to come that will try to test full functionality of the system while making sure integration tests _should_ all pass. The only test we've yet skipped here is the "accrue lots of interest" test that will likely require a specially compiled Compound Chain to manually set timestamps, and as such, we leave that test for a later date.

PR #218, opened 2021-03-04
Lock To Any [1.3.1]

hayesgm:
This patch adds "Lock to Any" which allows a user to lock an asset destined for either his/her own account or for an account from a different chain, per se. As Eth is the only current chain, this is mostly future proofing.

We also take this time to explore how to perform a safe upgrade. Specifically, the Starport events change after the Starport is upgraded. As such, we will upgrade the chain to a version (1.3.1) that will parse the old and new events from the Starport. Then we will upgrade the Starport. Finally, we will in a future version remove support for the old events.

We actually test that this upgrade pattern works in a new integration test. See `integration/__tests__/upgrade_to_131_scen.js`

Note: this unreverts "Lock to Any" and subsumes #210


integration/__tests__/lock_test.js
```diff
@@ -1,80 +0,0 @@
-const {
```
hayesgm, line 1: This is very out-dated compared to the new scenarios that it provides little to no value.

integration/__tests__/upgrade_to_131_scen.js
```diff
@@ -0,0 +1,68 @@
...
+const { decodeCall, getEventData } = require('../util/substrate');
+const { bytes32 } = require('../util/util');
+
+let scen_info = {
+  tokens: [
+    { token: "zrx", balances: { ashley: 1000 } }
+  ],
+};
+
+buildScenarios('Upgrade to 1.3.1', scen_info, [
```
hayesgm, line 13: This test is the crux of showing that upgrades will be safe.

integration/util/download.js
```diff
@@ -19,24 +19,46 @@ function download(url, path, options = {}, handle = null) {
     ...options,
   };
   return new Promise(async (resolve, reject) => {
-    let f = handle || await fs.open(path, 'w');
     let bail = false;
+    let f_;
```
hayesgm, line 23: Here we properly handle the case where we get a 404 (we used to write the 404 message to the file)

integration/util/scenario/eth.js
```diff
@@ -5,13 +5,15 @@ const { readContractsFile, deployContract, getContractAt } = require('../ethereu
 const { genPort } = require('../util');
 
 class Eth {
-  constructor(ethInfo, web3, web3Url, accounts, ganacheServer, ctx) {
+  // TODO: Do the construct
+  constructor(ethInfo, web3, web3Url, accounts, ganacheServer, version, ctx) {
```
hayesgm, line 9: Solidity contracts are now versioned so we can deploy older contracts for upgrade and compatibility tests.

integration/util/substrate.js
```diff
@@ -42,6 +43,11 @@ function waitForEvent(api, pallet, method, opts = {}) {
 
   let resolve, reject;
   let promise = new Promise((resolve_, reject_) => {
+    if (opts.timeout) {
```
hayesgm, line 46: This timeout should be helpful in stopping "forever running" tests.

types.json
```diff
@@ -12,7 +12,7 @@
   "AssetQuantity": "Quantity",
   "CashQuantity": "Quantity",
   "USDQuantity": "Quantity",
-  "LiquidityFactor": "Factor",
+  "LiquidityFactor": "(u128)",
```
coburncoburn, line 15: LiquidityFactor in here 2x?

types.json
```diff
@@ -414,6 +414,7 @@
     }
   },
   "Source10": "AccountRS",
+  "LiquidityFactor": "u128",
```
coburncoburn, line 417: other liquidity factor (u128) or u128?

coburncoburn (approved):
probable error in types.json

integration/util/substrate.js
waynenilsen: yes this helped me a lot thank you

pallets/cash/src/internal/events.rs
```diff
@@ -108,6 +108,13 @@ pub fn receive_event<T: Config>(
     event: ChainLogEvent,
     signature: ValidatorSig,
 ) -> Result<(), Reason> {
+    log!(
```
waynenilsen, line 111: Probably too chatty (?) not sure if we want to address this in some broader way though where during sync you don't show all the logs and you start showing them when you catch up.

types.json
waynenilsen, line 15: i don't think this is valid json due to reused key in the same map `LiquidityFactor`

pallets/cash/src/internal/lock.rs
```diff
@@ -0,0 +1,199 @@
...
+use crate::{
+    chains::ChainAccount,
+    core::{repay_and_supply_principal, sub_principal_amounts},
+    reason::Reason,
+    types::{CashIndex, CashPrincipalAmount},
+    CashPrincipals, ChainCashPrincipals, Config, Event, GlobalCashIndex, Module,
+    TotalCashPrincipal,
+};
+
+pub fn lock_cash_principal_internal<T: Config>(
```
waynenilsen, line 14: note to other reviewers, this function diff is really something like this afaict 
image

waynenilsen (approved):
Looks good 👍 

One tiny nit; I think we might want `LockV1` and `LockV2` rather than `LockOld` and `Lock` but that is really just a desire to avoid the potential dreaded `OldOld`.

~I'm not sure it is worth doing this as an integration test (or maybe this is what the upgrade test does) but I think we should "try out" the upgrade process locally with lock events before upgrade and after including sync from 0.~

The point of this PR was to add `integration/__tests__/upgrade_to_131_scen.js` the rest was already approved. From that perspective, I ran the test and read through it. It makes sense to me and seems to prove out the upgrade path.

pallets/cash/src/internal/events.rs
hayesgm: Yeah, I think we'll remove many of them once we get things stabilized.

types.json
hayesgm: I just reverted these changes, thanks.

PR #222, opened 2021-03-05
Allow Excess Signatures

hayesgm:
This patch is to allow the Starport to accept excess signatures-- that is, if you provide sigantures that aren't valid (either by duplication or passing in invalid signers, we do not strictly reject the notice. We simply do not count such entries and then tally the total at the end. This is to reduce frictions for users, esp. when the authorities are rotated before a notice is mined but it _might_ still have a sufficient quorum with the new authority set.

Note: should we still fail on duplicates?
Note: this is part of 1.4.1 which will improve Notices generally.

### DIFF
--- a/ethereum/contracts/Starport.sol
+++ b/ethereum/contracts/Starport.sol
@@ -370,14 +370,19 @@ contract Starport {
         bytes[] calldata signatures
     ) internal pure {
         address[] memory sigs = new address[](signatures.length);
+        uint sigsLen = 0;
+        uint authorityCount = authorities_.length;
+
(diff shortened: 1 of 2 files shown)


coburncoburn (approved):
seems good, not more expensive since doing the same checks already 👍

jflatow (approved):
I was wondering about why we need to allow duplicates but saw your comment and should be fine, why not...although its weird to think of a case where you would ever want to do that

hayesgm:
Yeah, it's mostly: we're trusting that we can probably ignore bad things, so let's just do that for all varieties of bad?

PR #223, opened 2021-03-05
Name Cash Token

hayesgm:
As we're public now, let's name the Cash Token. We can push this change as part of 1.4.1 to deploy the new implementation for Cash Token, unless we want to otherwise roll other Cash Token changes before we do this.

Note: we remove a variety of `// XXX` comments as they have, by and large, been addressed already.


rleshner (approved):
Easy!

PR #225, opened 2021-03-05
Debugging and fixes

hayesgm:


integration/util/scenario/validator.js
```diff
@@ -221,7 +221,7 @@ class Validators {
 
   async addValidator(name, validatorInfo) {
     let newValidator = buildValidator(name, validatorInfo, this.ctx);
-    await this.ctx.api().rpc.system.addReservedPeer(newValidator.asPeer());
+    await Promise.all(this.all().map((validator) => validator.api.rpc.system.addReservedPeer(newValidator.asPeer())));
```
maxwolff, line 224: ah nice i missed this, maybe this is why i got those panics

PR #227, opened 2021-03-05
Improve Notice Signing Mechanics

hayesgm:
This patch fixes a few items with notice signature collection. First, we make sure to check that signers are current validators before appending signatures to the notice pending set. Secondly, we fix the `recovery` code to fix the check for whether or not we _should_ sign (it was recovering to the wrong address and so the OCW were trying to re-sign every signature every block). Finally, we add full test coverage for this sub-module.


pallets/cash/src/chains.rs
```diff
@@ -386,10 +391,14 @@ impl Chain for Ethereum {
         hash
     }
 
-    fn recover_address(data: &[u8], signature: Self::Signature) -> Result<Self::Address, Reason> {
+    fn recover_user_address(data: &[u8], signature: Self::Signature) -> Result<Self::Address, Reason> {
         Ok(compound_crypto::eth_recover(data, &signature, true)?)
     }
 
+    fn recover_address(data: &[u8], signature: Self::Signature) -> Result<Self::Address, Reason> {
```
maxwolff, line 398: the same as above?

pallets/cash/src/internal/notices.rs
```diff
@@ -89,17 +95,11 @@ pub fn publish_signature(
                 _ => Err(Reason::SignatureMismatch),
             }?;
 
-            // require(signer == known validator); // XXX
-
-            // XXX sets?
-            // let signers_new = {signer | signers};
-            // let signatures_new = {signature | signatures };
+            // Note: we currently iterate all potentially all validators to check validity
+            if !Validators::iter().any(|(_,v)| ChainAccount::Eth(v.eth_address) == signer) {
```
maxwolff, line 99: so we do the sigs first and then check if the signer is a validator? kid of weird but :shrug

maxwolff (approved):

pallets/cash/src/chains.rs
hayesgm: The `bool` flag indicates whether or not the signature comes from a user. That is: was the signature `sign(data)` or `sign("\x19Ethereum Signed Message:" + data.length.toString() + data")`. This was the problem that caused us to keep double signing since we recovered the wrong version in our code that checked if we already signed.

pallets/cash/src/internal/notices.rs
hayesgm: good question on ordering, but I think either way works. We can optimize between the two.

PR #233, opened 2021-03-07
Rename Compound Chain to Gateway

hayesgm:
This patch renames `Compound Chain` to Gateway throughout the repo. We expect the name of the executable, etc will change and rename `compound-crypto` to `gateway-crypto` to match the new naming scheme. We may decide to keep a few select items named "compound chain" such as the name of terraform resources, etc, that might be too much effort to change, but that can be discussed in the conversation around this PR.


jflatow (approved):
Huge 💪

PR #235, opened 2021-03-09
Add Better Unsigned Validation

hayesgm:
This patch adds better validation to unsigned transactions. Specifically, we perform some basic checks that the extrinsics are likely to succeed (e.g. that they are signed by a current validator or the user signature is valid). We add complete test coverage of these functions, as well.

### DIFF
--- a/pallets/cash/src/internal/validate_trx.rs
+++ b/pallets/cash/src/internal/validate_trx.rs
@@ -2,6 +2,7 @@ use crate::{
     chains::{Chain, ChainAccount, Ethereum},
     internal,
     notices::EncodeNotice,
+    reason::Reason,
     AllowedNextCodeHash, Call, Config, Notices, Validators,
 };
 use codec::Encode;
@@ -11,32 +12,35 @@ use sp_runtime::transaction_validity::{TransactionSource, TransactionValidity, V
 
 #[derive(Eq, PartialEq, RuntimeDebug)]
 pub enum ValidationError {
+    InvalidInternalOnly,
     InvalidNextCode,
     InvalidValidator,
     InvalidSignature,
     InvalidCall,
     InvalidPriceSignature,
+    InvalidPrice(Reason),
     UnknownNotice,
 }
 
 pub fn validate_unsigned<T: Config>(
-    _source: TransactionSource,
+    source: TransactionSource,
     call: &Call<T>,
 ) -> Result<TransactionValidity, ValidationError> {
     match call {
-        Call::set_miner(_miner) => Ok(ValidTransaction::with_tag_prefix("Gateway::set_miner")
-            .priority(1)
-            .longevity(10)
-            .propagate(true)
-            .build()),
+        Call::set_miner(_miner) => match source {
+            TransactionSource::Local => Ok(ValidTransaction::with_tag_prefix("Gateway::set_miner")
+                .longevity(1)
+                .build()),
+            _ => Err(ValidationError::InvalidInternalOnly),
+        },
         Call::set_next_code_via_hash(next_code) => {
             let hash = <Ethereum as Chain>::hash_bytes(&next_code);
 
             if AllowedNextCodeHash::get() == Some(hash) {
                 Ok(
                     ValidTransaction::with_tag_prefix("Gateway::set_next_code_via_hash")
-                        .priority(1)
-                        .longevity(10)
+                        .priority(100)
+                        .longevity(32)
                         .and_provides(hash)
                         .propagate(true)
                         .build(),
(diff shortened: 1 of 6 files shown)


pallets/cash/src/internal/validate_trx.rs
```diff
@@ -0,0 +1,570 @@
...
+}
+
+pub fn validate_unsigned<T: Config>(
+    _source: TransactionSource,
+    call: &Call<T>,
+) -> Result<TransactionValidity, ValidationError> {
+    match call {
+        Call::set_miner(_miner) => Ok(ValidTransaction::with_tag_prefix("Gateway::set_miner")
+            .priority(1)
+            .longevity(10)
```
coburncoburn: surprisign to see set_miner here, isn't it set in inherent? Longevity also seems weird, as if the tx's could float around in the pool and maybe it doesn't get mined the next bvlock and someone gets 2 rewards in a row for authoring one block?

pallets/cash/src/internal/validate_trx.rs
```diff
@@ -0,0 +1,570 @@
...
+                .priority(3)
+                .longevity(10)
+                .and_provides((sender, nonce))
+                .and_requires((sender, nonce - 1))
+                .propagate(true)
+                .build()),
+            }
+        }
+        Call::post_price(payload, signature) => {
+            if internal::oracle::check_signature::<T>(&payload, &signature) == Ok(true) {
```
coburncoburn, line 98: maybe also validate that it is more recent than current price?

pallets/cash/src/internal/validate_trx.rs
hayesgm: I think we should actually check `TransactionSource` is internal, maybe with a few others, as well.

pallets/cash/src/internal/validate_trx.rs
hayesgm: I do want to try to keep these semi-lightweight and then up them as need be. E.g. that would be yet another read here, but I agree, someone might just spam with out-of-date messages. I think in some of these cases, we should check deeper if the source is _not_ Internal.

pallets/cash/src/internal/validate_trx.rs
hayesgm: I updated with better usage of transaction source here.

pallets/cash/src/internal/validate_trx.rs
hayesgm: I updated to do this check when this is not coming from internal

hayesgm:
Okay, this now has a significant number of improvements and fixes to the original validation code. Also, I've tested a few scenarios and local dev and everything seems to work pretty nominally.

pallets/cash/src/internal/validate_trx.rs
```diff
@@ -0,0 +1,637 @@
...
+                match source {
+                    TransactionSource::Local => {
+                        Ok(ValidTransaction::with_tag_prefix("Gateway::post_price")
+                            .priority(100)
+                            .and_provides(signature)
```
jflatow, line 103: Shouldn't this be the only one we need and propagate(false)?

hayesgm: Post Price is allowed to be called externally, in case OCW aren't sufficiently updating prices.

waynenilsen (commented):
What are the units of priority? Why is 100 a better value than what was there 1,2,3,4,5?

Perhaps post_price should have longevity 1 ?

jflatow (approved):

pallets/cash/src/internal/validate_trx.rs
```diff
@@ -0,0 +1,637 @@
...
+
+pub fn validate_unsigned<T: Config>(
+    source: TransactionSource,
+    call: &Call<T>,
+) -> Result<TransactionValidity, ValidationError> {
+    match call {
+        Call::set_miner(_miner) => match source {
+            TransactionSource::Local => Ok(ValidTransaction::with_tag_prefix("Gateway::set_miner")
+                .longevity(1)
+                .build()),
```
toninorair, line 33: if the intention was `propagate(false)` here, it should be explicitly set

hayesgm:
> What are the units of priority? Why is 100 a better value than what was there 1,2,3,4,5?
> 
> Perhaps post_price should have longevity 1 ?

I just feel 100 is neutral. No reason, but if you choose "1" then you can't choose "a number between 1 and 0" so 100 gives you space if you want to adjust the priority (e.g. you can order something by making it 99, which we don't do yet).

toninorair (approved):

toninorair (approved):

PR #237, opened 2021-03-09
Add Oracle Pallet

hayesgm:
This patch makes Oracle its own pallet, which can therefore define its own Inherents. Also, this actually cleans up a lot of separation of concerns since many of the aspects of the Oracle should be generally opaque to the Cash pallet code. This is still a WIP.


jflatow (approved):
This is the refactor of my dreams, the only thing I would change is that I still think its nice to refer to USD.ticker and not split into a const USD in 1 file and const USD_TICKER in another (we could still have the `ticker.rs` but just the consts defined once). But otherwise this is really awesome

pallets/oracle/src/inherent.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl<T: Config> ProvideInherent for Module<T> {
+    type Call = Call<T>;
+    type Error = InherentError;
+    const INHERENT_IDENTIFIER: InherentIdentifier = INHERENT_IDENTIFIER;
+
+    fn create_inherent(data: &InherentData) -> Option<Self::Call> {
+        match extract_inherent_data(data) {
+            Ok(miner) => Some(Call::set_miner(miner)),
```
toninorair: should this miner part be here or it is only inherent for prices?

toninorair (approved):

jflatow (approved):
🥇

pallets/oracle/src/inherent.rs
hayesgm: Yup, this was older code, now fixed.

PR #238, opened 2021-03-10
notes on manual hack of building chainspec

coburncoburn:
this process is half baked so far with some manual intervention, adding some notes for future progress


hayesgm:
These hacks were merged into the official scripts. Thanks for documenting.

PR #239, opened 2021-03-10
Make eth_recover calls via runtime_interface for speed

waynenilsen:


jflatow (approved):
Awesome, do we observe a massive speedup using this? I agree it would be great to get this into m5

waynenilsen:
> Awesome, do we observe a massive speedup using this? 

Not sure to be honest, didn't benchmark it

PR #240, opened 2021-03-10
Make dispatching notices non-fallible

jflatow:
Still need to figure out how to fix the holds thing


pallets/cash/src/internal/notices.rs
```diff
@@ -4,13 +4,174 @@ use frame_support::storage::{
...
+                (ChainAccount::Eth(eth_account), ChainHash::Eth(eth_parent_hash)) => {
+                    CashExtractionNotice::Eth {
+                        id: notice_id,
+                        parent: eth_parent_hash,
+                        account: eth_account,
+                        principal: principal.0,
+                    }
+                }
+
+                _ => panic!("XXX not implemented"), // generate these w/ macros?
```
hayesgm, line 70: So just checking, we basically just took these from failable into panicing? I get that we can probably reason this, but it's theoretically possible that the ADTs don't match, right? I was hoping we could move the checks into the inputs somehow to make it strictly impossible or the have failed prior to being in "effects" territory.

pallets/cash/src/lib.rs
```diff
@@ -396,22 +388,23 @@ fn has_requisite_signatures(notice_state: NoticeState, validators: &Vec<Validato
     }
 }
 
+// XXX should this have all these printlns?
```
hayesgm, line 391: They are actually very useful right now, which is why they have been kept in. This patch needs to make the corresponding change in types.json.

pallets/cash/src/reason.rs
```diff
@@ -50,8 +50,8 @@ pub enum Reason {
     RatesError(RatesError),
     RepayTooMuch,
     SelfTransfer,
-    SetYieldNextError(SetYieldNextError),
     SerdeError,
+    SetYieldNextError(SetYieldNextError),
```
hayesgm, line 54: We should really stop alphabetizing this, since it makes types.json hard to stay consistent.

hayesgm (commented):
This is a lot cleaner. My only question is around changing from "failable" to "panics." I'm just not entirely convinced that's a cleaner overall state.

PR #242, opened 2021-03-11
[COMP-1227] Long Compounding Interest

waynenilsen:
This patch is to update Gateway to use long compounding interest (as opposed to interest realized every block). This is in line with the new specification of On Initialize. Crucially, this means that Cash Token and Gateway will have identical representations of interest and we should not need to sync in a routine manner. We should carefully unit test this change.

* updated on_initialize function in `core.rs`
* unit tests focusing on yield transition

### DIFF
--- a/pallets/cash/src/core.rs
+++ b/pallets/cash/src/core.rs
@@ -34,9 +34,9 @@ use crate::{
         Timestamp, USDQuantity, Units, ValidatorIdentity, CASH,
     },
     AssetBalances, AssetsWithNonZeroBalance, BorrowIndices, CashPrincipals, CashYield,
-    CashYieldNext, ChainCashPrincipals, Config, Event, GlobalCashIndex, LastIndices,
-    LastMinerSharePrincipal, LastYieldTimestamp, Miner, Module, SupplyIndices, SupportedAssets,
-    TotalBorrowAssets, TotalCashPrincipal, TotalSupplyAssets,
+    CashYieldNext, ChainCashPrincipals, Config, Event, GlobalCashIndex, LastBlockTimestamp,
+    LastIndices, LastMinerSharePrincipal, LastYieldCashIndex, LastYieldTimestamp, Miner, Module,
+    SupplyIndices, SupportedAssets, TotalBorrowAssets, TotalCashPrincipal, TotalSupplyAssets,
 };
 
 #[macro_export]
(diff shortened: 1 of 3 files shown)


pallets/cash/src/core.rs
```diff
@@ -1226,20 +1231,31 @@ pub fn on_initialize_internal<T: Config>(
     if now == 0 {
         return Err(Reason::TimeTravelNotAllowed);
     }
-    if last_yield_timestamp == 0 {
-        // this is the first time we have seen a valid time, set it
-        LastYieldTimestamp::put(now);
+    if last_yield_timestamp == 0 || last_block_timestamp == 0 {
```
waynenilsen, line 1238: dev note - this is something of a belt and suspenders issue - last_yield_timestamp is actually meant to be set by `config()` / `chainSpec` but having a safegaurd here may not be the worst thing? It was here before, i just updated it

pallets/cash/src/lib.rs
```diff
@@ -191,6 +193,12 @@ decl_storage! {
 
         /// Validator spread due to miner of last block.
         LastMinerSharePrincipal get(fn last_miner_share_principal): CashPrincipalAmount;
+
+        /// The timestamp of the previous block or defaults to timestamp at genesis.
+        LastBlockTimestamp get(fn last_block_timestamp): Timestamp;
```
waynenilsen, line 198: dev note - not sure if we want `config()` here and/or how disruptive that is for the time being

pallets/cash/src/lib.rs
```diff
@@ -191,6 +193,12 @@ decl_storage! {
...
+
+        /// The cash index of the previous yield accrual point or defaults to initial cash index.
+        LastYieldCashIndex get(fn last_yield_cash_index): CashIndex;
```
waynenilsen, line 201: I think using Default::default() = cash index of 1 for now makes sense and we don't need `config()` or `add_extra_genisis` for this right now.

pallets/cash/src/core.rs
jflatow: Never heard that term before, I like it ;) Not sure about the right thing here overall yet but I think this is good now

pallets/cash/src/lib.rs
jflatow: Yea good call out, this is the question

jflatow (approved):

PR #247, opened 2021-03-12
Rpcs for asset and cash data

mykelp:
This patch adds two rpcs, one for retrieving all necessary asset data for a given account (i.e. balance, supply rate, borrow rate, price, liquidity factor), and another for retrieving cash data for a given account (i.e. balance, cash_yield, price).


jflatow (approved):
:clap:

hayesgm (approved):

PR #250, opened 2021-03-15
Clean up `receive_event` signing and signature recovery

toninorair:
This pr was inspired by a discussion in the discord channel regarding non-standard signature signing and verification for the `receive_event` method. The proposed approach makes it match other code parts better. At first, I tried passing `ChainSignature`, but it was messy to recover and match the chain later, so just kept the Ethereum signature type.


hayesgm (approved):
Awesome, looks good to me.

jflatow (approved):

PR #251, opened 2021-03-15
Only auth notices initiate hold

maxwolff:
only change auth notices should initiate hold, and only change_auth extrinsic should require there to be no hold (other era changing extrinsics should not fail)

### DIFF
--- a/pallets/cash/src/internal/change_validators.rs
+++ b/pallets/cash/src/internal/change_validators.rs
@@ -2,10 +2,12 @@ use frame_support::storage::{IterableStorageMap, StorageMap};
 
 use crate::{
     internal, reason::Reason, require, types::ValidatorKeys, Config, Event, Module, NextValidators,
-    SessionInterface,
+    NoticeHolds, SessionInterface,
 };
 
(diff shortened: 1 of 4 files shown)


pallets/cash/src/internal/change_validators.rs
```diff
@@ -17,7 +19,6 @@ pub fn change_validators<T: Config>(validators: Vec<ValidatorKeys>) -> Result<()
         NextValidators::take(&id);
     }
     for keys in &validators {
-        NextValidators::take(&keys.substrate_id);
```
maxwolff, line 20: redundant w above line

hayesgm (commented):
Where do we build the notice hold object in this patch? I just don't see it in the diff.

maxwolff:
> Where do we build the notice hold object in this patch? I just don't see it in the diff.

hey sorry jumped the gun w the pr before i pushed the last commit, its here https://github.com/compound-finance/gateway/pull/251/files#diff-7e313ecc9237df84e3bb0c61f2a3199473f1c954769bce9453cdba30257270e4R165

hayesgm (approved):
Looks great, we can release this in m7. We should probably start to add unit tests to the extrinsic itself.

jflatow (approved):

PR #253, opened 2021-03-16
Switch RPC to use String types where necessary

waynenilsen:


integration/util/scenario/validator.js
```diff
@@ -166,9 +188,23 @@ class Validator {
     });
 
     const wsProvider = new WsProvider(`ws://localhost:${this.wsPort}`);
+    const types = await loadTypes(this.ctx);
     const api = await ApiPromise.create({
       provider: wsProvider,
-      types: await loadTypes(this.ctx)
+      types: {
+        ...types,
+        ApiAssetData: {
```
waynenilsen: dev note - not sure about this one .. just matching castle here

torreyatcitty (approved):
lgtm

integration/util/scenario/validator.js
mykelp: removed.

jflatow (approved):
💪

PR #254, opened 2021-03-16
Changes after the OZ audit

toninorair:
Changes for this report https://gist.github.com/jcarpanelli/530f48abbbe89fde8b48beaadc6804eb

### DIFF
--- a/ethereum/contracts/Starport.sol
+++ b/ethereum/contracts/Starport.sol
@@ -31,7 +31,6 @@ contract Starport {
     event Unlock(address indexed account, uint amount, address asset);
     event UnlockCash(address indexed account, uint amount, uint128 principal);
     event ChangeAuthorities(address[] newAuthorities);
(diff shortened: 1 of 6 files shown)


ethereum/contracts/Starport.sol
```diff
@@ -31,7 +31,6 @@ contract Starport {
     event Unlock(address indexed account, uint amount, address asset);
     event UnlockCash(address indexed account, uint amount, uint128 principal);
     event ChangeAuthorities(address[] newAuthorities);
-    event SetFutureYield(uint128 nextCashYield, uint128 nextCashYieldIndex, uint nextCashYieldStart);
```
hayesgm: I'm not sure we need to remove this. I think it's fine if both the Starport and Cash Token emit such an event.

hayesgm (commented):
Looks good to me, do we need any additional unit tests for these changes? Also, not sure we need to remove `SetFutureYield` from Starport.

jflatow (approved):

PR #255, opened 2021-03-17
Test Future Yield

hayesgm:
This patch tests setting the future yield and ensures that we are, in fact, able to properly sign and accept the notice in the Starport. Types.json misconfigured the type and thus we needed to patch the type for this fix.

### DIFF
--- a/types.json
+++ b/types.json
@@ -153,8 +153,8 @@
     "id": "NoticeId",
     "parent": "[u8; 32]",
     "next_cash_yield": "u128",
-    "next_cash_yield_start_at": "u128",
-    "next_cash_index": "u128"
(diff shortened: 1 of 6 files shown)


jflatow (approved):

types.json
```diff
@@ -153,8 +153,8 @@
     "id": "NoticeId",
     "parent": "[u8; 32]",
     "next_cash_yield": "u128",
-    "next_cash_yield_start_at": "u128",
-    "next_cash_index": "u128"
+    "next_cash_index": "u128",
+    "next_cash_yield_start_at": "u64"
```
jflatow, line 157: Ouch, we really need to generate this

PR #261, opened 2021-03-18
Unit Test Change Authority Extrinsic

maxwolff:


hayesgm (approved):
Looks great. The only other thing we might want to test is that it clears the existing `NextValidators`, right?

maxwolff:
> Looks great. The only other thing we might want to test is that it clears the existing `NextValidators`, right?

yea good catch, added

PR #264, opened 2021-03-19
CLI Configuration

waynenilsen:
We were previously inserting a lot of the validator configuration via environment variables only.

We now have the ability to configure the node via CLI with an environment variable override. This allows everyone that already set up environment variables to continue using them while at the same time allowing our integration tests to set the variables in the CLI.

### DIFF
--- a/pallets/runtime-interfaces/src/lib.rs
+++ b/pallets/runtime-interfaces/src/lib.rs
@@ -11,6 +11,14 @@ pub struct Config {
     eth_starport_address: Vec<u8>,
 }
 
+#[derive(Clone)]
+pub struct ValidatorConfig {
+    eth_key_id: String,
+    eth_rpc_url: String,
+    miner: String,
+    opf_url: String,
+}
+
 /// XXX Possible sanity checks for config fields here
 impl Config {
     pub fn update(&mut self, new: Config) {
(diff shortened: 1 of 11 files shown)


pallets/runtime-interfaces/src/lib.rs
```diff
@@ -58,12 +86,14 @@ pub trait ConfigInterface {
...
-const ETH_KEY_ID_ENV_VAR_DEV_DEFAULT: &str = gateway_crypto::ETH_KEY_ID_ENV_VAR_DEV_DEFAULT;
 const ETH_RPC_URL_ENV_VAR: &str = "ETH_RPC_URL";
 const MINER_ENV_VAR: &str = "MINER";
-const ETH_RPC_URL_ENV_VAR_DEV_DEFAULT: &str = "https://goerli-eth.compound.finance";
 const OPF_URL_ENV_VAR: &str = "OPF_URL";
-const OPF_URL_ENV_VAR_DEFAULT: &str = "https://prices.compound.finance/coinbase";
+
+const ETH_KEY_ID_DEFAULT: &str = gateway_crypto::ETH_KEY_ID_ENV_VAR_DEV_DEFAULT;
+const MINER_DEFAULT: &str = "Eth:0x0000000000000000000000000000000000000000";
+const ETH_RPC_URL_DEFAULT: &str = "https://goerli-eth.compound.finance";
```
hayesgm: Should probably change to Ropsten or maybe just not at all?

hayesgm (approved):
excellent. excellent. excellent.

PR #265, opened 2021-03-19
Show the Dangers of a Re-org

hayesgm:
This patch adds a test that currently shows the danger of our a re-org. We will use this as one of the bases to test that our new work on re-org resistance is effective in real-life scenarios.

### DIFF
--- a/integration/__tests__/reorg_scen.js
+++ b/integration/__tests__/reorg_scen.js
@@ -34,5 +34,41 @@ buildScenarios('Chain Re-organization Scenarios', lock_scen_info, [
       expect(await starport.tokenBalance(usdc)).toEqual(0);
       expect(await bert.tokenBalance(usdc)).toEqual(100);
     }
+  },
+  {
+    only: true,
+    name: 'Re-org with Identical Tx',
+    scenario: async ({ ashley, bert, ether, chain, snapshot, starport, eth, sleep }) => {
+      let crazyLock = await eth.__deploy('CrazyLock', [starport.ethAddress()]);
+      let now = Date.now();
+      let nowEven = now - (now % 2);
+      let nowOdd = nowEven + 1;
+
+      let snapshotId = await eth.snapshot();
+      await eth.mine(1, nowEven);
+      let tx0 = await crazyLock.methods.crazyLock(ashley.ethAddress()).send({value: 0.1e18, from: ashley.ethAddress()});
+      await eth.mine(20);
+      await chain.waitForEthProcessEvent('cash', ether.lockEventName());
+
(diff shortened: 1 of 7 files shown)


waynenilsen (approved):
very nice, working well.

I wonder if we want to write these up so they're failing somehow today since this is ..... not expected behavior or behavior that we would like to correct/change? In the spirit of TDD something to the effect of "step 1 write a failing test". Feel free to leave it, either way we will need to refactor just seems odd to see ✅ on "reorg allows double spend" expectation lol.

integration/__tests__/reorg_scen.js
```diff
@@ -0,0 +1,74 @@
...
+
+      // Now it's time for the re-org
+      await eth.restore(snapshotId);
+      await ashley.tokenTransfer(bert, 100, usdc);
+      await eth.mine(20);
+
+      // Uh-oh
+      expect(await ashley.tokenBalance(usdc)).toEqual(0);
+      expect(await ashley.chainBalance(usdc)).toEqual(100);
+      expect(await starport.tokenBalance(usdc)).toEqual(0);
```
jflatow, line 35: 💪

integration/util/scenario/chain.js
```diff
@@ -181,6 +181,12 @@ class Chain {
     });
   }
 
+  async tokenBalance(token, chainAccount) {
+    console.log(["gghy88", token.toChainAsset(), chainAccount]);
```
jflatow: ?

jflatow (approved):
Very impressive, one question

jflatow:
> very nice, working well.
> 
> I wonder if we want to write these up so they're failing somehow today since this is ..... not expected behavior or behavior that we would like to correct/change? In the spirit of TDD something to the effect of "step 1 write a failing test". Feel free to leave it, either way we will need to refactor just seems odd to see ✅ on "reorg allows double spend" expectation lol.

No matter what this is going to be the case right, and this test should pass (after re-org, starport doesn't have assets and transfer succeeds)? I think the breaking test we would want is that after the re-org, the Gateway state reflects the new reality, rather than continuing to debit/credit the funds

integration/util/scenario/chain.js
hayesgm: FWIW, there are so many logs in integration tests that I use random strings so I can easily "find" a message in the clutter 🤦

hayesgm:
I'll mark these as `skip` right now. I'd rather them show the current behavior because it's clearly wrong. If we just checked that they don't double-account, then it might have been the case that it was a set-up bug, etc, if they were green.

hayesgm:
Also, depending on how we "clawback" the actual expectations aren't clearly defined yet.

PR #275, opened 2021-04-08
Start to Improve Scenarios as Regression Tests

hayesgm:
This patch starts to improve scenarios so that they can pass consistently. We make a few changes:

1. We add a `QUIET_SCENARIOS` flag that pushes all scenario output to temp files. This makes it possible to run scenarios and see the output and not 1,000 pages of validator text.
2. We add an `EventTracker` module that handles tracking events without using global vars (which was probably messing up test cases). Events are hard because they come in from a subscription in PolkadotJS, but our scenarios often ask "Have you seen any Substrate `Lock()` events?" and we have to say something like "Yes, we saw one more recently than the last time you asked." That can be a bit tricky and now that state is encapsulated and doesn't leak from one scenario to the next (this may have been a large bug in running sequential tests).
3. We start to move back away from `--runInBand` which means tests will run in separate processes and concurrently again. We may want to make it easier to not run in band for building tests (it's just a matter of passing the flag, but engineers might forget/not know how to do this, so we should make that clear).
4. We remove `only: true` and start to add back multiple tests so we can measure what is working and not. So far the results are not too excessively shabby.

### DIFF
--- a/integration/__tests__/gov_scen.js
+++ b/integration/__tests__/gov_scen.js
@@ -12,7 +12,7 @@ let gov_scen_info = {
 buildScenarios('Gov Scenarios', gov_scen_info, [
   {
     name: "Update Interest Rate Model by Governance",
-    scenario: async ({ ctx, zrx, chain, starport, sleep }) => {
+    scenario: async ({ ctx, zrx, chain, starport }) => {
       let newKink = {
         Kink: {
           zero_rate: 100,
@@ -38,7 +38,7 @@ buildScenarios('Gov Scenarios', gov_scen_info, [
         }
       },
     },
-    scenario: async ({ ctx, zrx, chain, starport, curr, sleep }) => {
+    scenario: async ({ ctx, zrx, chain, starport, curr }) => {
       expect(await chain.getSemVer()).toEqual([1, 2, 1]);
       let currHash = await curr.hash();
       let extrinsic = ctx.api().tx.cash.allowNextCodeWithHash(currHash);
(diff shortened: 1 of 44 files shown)


jflatow (approved):
🆒

integration/__tests__/transfer_scen.js
```diff
@@ -38,19 +38,14 @@ buildScenarios('Transfer Scenarios', transfer_scen_info, { beforeEach: lockUSDC
     }
   },
   {
-    only: true,
+    skip: true,
```
jflatow, line 41: Why `skip`?

hayesgm: It's an expected failure at this point. I am removing `only` tags, but trying to get 100% of non-skipped tests passing.

integration/__tests__/gov_scen.js
```diff
@@ -21,48 +19,28 @@ buildScenarios('Gov Scenarios', gov_scen_info, [
...
         }
       };
-      let extrinsic = ctx.api().tx.cash.setRateModel(zrx.toChainAsset(), newKink);
+      let extrinsic = api.tx.cash.setRateModel(zrx.toChainAsset(), newKink);
       await starport.executeProposal("Update ZRX Interest Rate Model", [extrinsic]);
       expect(await chain.interestRateModel(zrx)).toEqual(newKink);
     }
   },
   {
-    name: "Upgrade Chain WASM [Set Code]",
```
hayesgm, line 30: This is removed since we removed `sudo`

PR #277, opened 2021-04-08
Add logging for validate_unsigned

toninorair:
`sp_tracing` didn't give many benefits (it can give better error tracing for tests for sure), and unclarity how to properly use it for the testnet or even if it needed, led me to take this simple way for missing `validate_unsigned` error logging.


coburncoburn (approved):
seems reasonable

PR #278, opened 2021-04-08
Cob/sequential trx requests

coburncoburn:
.exec (My Trx Request) added to repl
sort trx requests in transaction pool

### DIFF
--- a/chains/scripts/README.md
+++ b/chains/scripts/README.md
@@ -1,0 +1,3 @@
+usage: scripts to be run in the context of the repl
+#in `yarn repl`
+> eval(fs.readFileSync("./scripts/${my_script}.js").toString())
--- a/chains/scripts/README.md
+++ b/chains/scripts/README.md
@@ -1,3 +1,2 @@
 usage: scripts to be run in the context of the repl
-#in `yarn repl`
-> eval(fs.readFileSync("./scripts/${my_script}.js").toString())
+yarn repl -s path_to_my_script.js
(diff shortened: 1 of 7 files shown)


chains/scripts/add_users.js
```diff
@@ -0,0 +1,50 @@
...
+async function feedAccount(n, r, accounts = [] ) {
+    const user = saddle.account
+    var accounts
+    var cash_nonce = n
+    if (r > 2000) return;
+
+    if (accounts.length == 0 || i == accounts.length) {
+        console.log(r)
+        accounts =  
+            await fetch(`https://api.compound.finance/api/v2/account?page_number=${r}&page_size=200`)
```
hayesgm, line 15: Wouldn't it just have been easier to do `0x000000000...1`, `0x000000000...2` and so on?

chains/scripts/README.md
```diff
@@ -0,0 +1,3 @@
+usage: scripts to be run in the context of the repl
```
hayesgm, line 1: I think we already have an option for this in the repl, if I recall it should be `yarn repl -s ./scripts/add_user.js`

hayesgm (approved):
Overall looks great.

chains/scripts/add_users.js
coburncoburn: yeah, maybe I'll change it up and do both so we get "realer" data and also more pathological data.

chains/scripts/README.md
coburncoburn: oh cool. is there one that keeps the shell open after running the script?

PR #280, opened 2021-04-15
Liquidation Scenarios

hayesgm:
This patch starts to build a full array of liquidation scenarios and starts to uncover some buggy behavior. There are still more tests to write, but we'll need to fix the specification in regards to close factor limitations and healthiness checks.


jflatow (commented):
Awesome stuff 👏

PR #281, opened 2021-04-15
Add weights for every extrinsic

maxwolff:
adds the rest of the benchmarks. dont anchor on the absolute weight numbers too hard, they are just being run locally in this pr, but the relative values should be roughly representative.


pallets/cash/src/internal/assets.rs
```diff
@@ -40,7 +40,7 @@ pub fn set_rate_model<T: Config>(
     Ok(())
 }
 
-pub fn support_asset<T: Config>(asset: ChainAsset, asset_info: AssetInfo) -> Result<(), Reason> {
-    SupportedAssets::insert(&asset, asset_info);
+pub fn support_asset<T: Config>(asset_info: AssetInfo) -> Result<(), Reason> {
```
maxwolff, line 42: asset is in asset_info, so redundant. anywhere else i need to make this change?

pallets/cash/src/lib.rs
```diff
@@ -494,42 +515,42 @@ decl_module! {
...
-        #[weight = (1, DispatchClass::Operational, Pays::No)] // XXX
+        #[weight = (<T as Config>::WeightInfo::allow_next_code_with_hash(), DispatchClass::Operational, Pays::No)] // XXX
         pub fn allow_next_code_with_hash(origin, hash: CodeHash) -> dispatch::DispatchResult {
             ensure_root(origin)?;
             Ok(check_failure::<T>(internal::next_code::allow_next_code_with_hash::<T>(hash))?)
         }
 
         /// Sets the allowed next code hash to the given hash. [User] [Free]
-        #[weight = (1, DispatchClass::Operational, Pays::No)] // XXX
+        #[weight = (<T as Config>::WeightInfo::set_next_code_via_hash(code.len().try_into().unwrap()), DispatchClass::Operational, Pays::No)] // XXX
```
maxwolff: is this safe? need to convert a usize into u32. not really sure what happens if it fails

node/Cargo.toml
```diff
@@ -94,7 +94,7 @@ sp-timestamp = { default-features = false, git = 'https://github.com/compound-fi
 frame-system = { git = 'https://github.com/compound-finance/substrate', branch = 'master' }
 
 [features]
-default = ['with-parity-db']
+default = ['with-rocks-db', 'runtime-benchmarks']
```
maxwolff: benchmarking w rocks and running parity is a bit of a mismatch, but had to do it to get it to run. apparently benchmarking on parity is still wip

hayesgm: I don't think these should be defaults, right? We should move `'with-rocks-db'` under `runtime-benchmarks = [` on line 100 and remove both of these from defaults.

pallets/cash/src/lib.rs
```diff
@@ -470,6 +473,29 @@ impl<T: Config> frame_support::traits::EstimateNextSessionRotation<T::BlockNumbe
     }
 }
 
+// weight is 0 if request is invalid
```
hayesgm: Is 0 the right default? Maybe make it semi-high just in case this comes up in the real world for some reason?

pallets/cash/src/lib.rs
hayesgm: Well this is 4000MB, which is... probably a reasonable constraint. Why do we need to condense it to 32-bits? Could we just have an arb. high weight for code lens > u32::MAX?

hayesgm (commented):
You got more weights than Gold's Gym. Looks great. Few minor comments.

node/Cargo.toml
maxwolff: yes, my mistake

pallets/cash/src/weights.rs
```diff
@@ -49,49 +51,156 @@ use sp_std::marker::PhantomData;
...
+    }
+    fn exec_trx_request_transfer() -> Weight {
+        (567_000_000 as Weight)
+            .saturating_add(RocksDbWeight::get().reads(19 as Weight))
+            .saturating_add(RocksDbWeight::get().writes(13 as Weight))
     }
-    fn receive_chain_reorg() -> Weight {
-        (243_000_000 as Weight) // XXX
+    fn exec_trx_request_liquidate() -> Weight {
+        (1_011_000_000 as Weight)
```
hayesgm, line 202: So bulkyyyy

hayesgm (approved):
lgtm, awesome work!

PR #282, opened 2021-04-16
Build Ethereum Contract to test Trx Request from Chain

hayesgm:
This patch adds a new contract (and a `TrxRequestBuilder`) which tests executing a trx request from on-chain. The builder is rudimentary, but can be the start of a full-time builder. The contract `Locker` just locks Eth and transfers it to you (the sender)'s account on Gateway. I believe this is currenty failing due to `(Transfer Cash Max 0x...)` failing right now.

Note: the one bug discovered so far is that events from a trx are processed in reverse order. We should ensure we fix that in our new event system.

### DIFF
--- a/integration/__tests__/ethereum_scen.js
+++ b/integration/__tests__/ethereum_scen.js
@@ -1,0 +1,18 @@
+const { buildScenarios } = require('../util/scenario');
+
+let ethereum_scen_info = {
+  tokens: []
+};
+
+buildScenarios('Ethereum Scenarios', ethereum_scen_info, [
+  {
+    name: "Locker: Lock and Extract",
+    scenario: async ({ chain, zrx, eth, starport }) => {
+      let locker = await eth.__deploy('Locker', [starport.ethAddress()]);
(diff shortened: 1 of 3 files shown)


jflatow (commented):
Looking good 💪

jflatow:
Hm what's the status of this now?

integration/__tests__/ethereum_scen.js
```diff
@@ -0,0 +1,18 @@
...
+  tokens: []
+};
+
+buildScenarios('Ethereum Scenarios', ethereum_scen_info, [
+  {
+    name: "Locker: Lock and Extract",
+    scenario: async ({ chain, zrx, eth, starport }) => {
+      let locker = await eth.__deploy('Locker', [starport.ethAddress()]);
+      await locker.methods.lockAndtransfer().send({ from: eth.root(), value: 0.1e18 });
+      // This should, if all goes well, XXX
```
jflatow, line 13: Kept just the contracts since the 'max transfer less fee' is already integrated. This scen is incomplete but going to merge so we don't lose these useful contracts

PR #283, opened 2021-04-18
Max Cash Less Fee

hayesgm:
This patch fixes "Transfer Max Cash" to subtract out the fee before calculating "max" so it doesn't necessarily push you into a negative cash position (since, e.g. if you had 5 CASH and then you transfered max, we would calculate that as "you want to transfer 5 Cash" and then charge you 0.01 Cash as a transfer fee and end up with a -0.01 CASH balance. Now we say you want to transfer 4.99 and you'll end up with exactly 0 CASH balance afterwards).
    
Additionally, we un-mock the tests and start to build a nice little suite of unit tests.

Finally, we also add support for partial debug compilation, which should speed up `cargo build` builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,6 @@
+[profile.dev]
+split-debuginfo = 'unpacked'
+
 [profile.release]
 panic = 'unwind'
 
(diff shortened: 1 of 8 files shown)


jflatow (approved):
LGTM 💪

PR #284, opened 2021-04-20
Liquidate Liquidity Check

hayesgm:
This patch adds a liquidity check to a user to ensure they are underwater before we allow liquidations on the account. There are still other areas to consider (i.e. close factor and collateral quantity), but this is a first start on fixing some bugs in the liquidation function.

Additionally, we add significant test coverage to the `has_liquidity_xxx` functions to ensure they are all individually test in a range of scenarios. We found no serious bugs in that process, but we should consider breaking on some invalid inputs (e.g. for unsupported assets, etc). These are currently left as TODOs.

### DIFF
--- a/pallets/cash/src/core.rs
+++ b/pallets/cash/src/core.rs
@@ -26,15 +26,16 @@ use crate::{
     chains::{ChainAccount, ChainAsset, ChainHash, ChainId},
     events::ChainLogEvent,
     factor::Factor,
-    internal, log,
+    internal::{self, liquidity},
+    log,
     params::{MIN_TX_VALUE, TRANSFER_FEE},
     portfolio::Portfolio,
     rates::APR,
     reason::{MathError, Reason},
     types::{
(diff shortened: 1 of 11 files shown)


pallets/cash/src/core.rs
```diff
@@ -734,7 +745,11 @@ pub fn liquidate_internal<T: Config>(
         )?;
 
     require!(
-        has_liquidity_to_reduce_asset_with_added_collateral::<T>(
+        !liquidity::has_non_negative_liquidity::<T>(borrower)?,
+        Reason::SufficientLiquidity
```
jflatow, line 749: 😹

jflatow (approved):
Looks great - thanks for isolating 💯

coburncoburn (approved):
based

pallets/cash/src/core.rs
hayesgm: I might change it, but it just feels so right.

PR #285, opened 2021-04-20
Handle Unpriced and Mismatched Assets

hayesgm:
To improve completeness, we add checks for matching asset prices to the assets themselves and ensure prices are set when checking liquidity. We change the behavior from pallet-oracle to return an optional price and use a default zero in the places where it's currently expected. We make no effort here to decide if _that_ is the correct behavior- just to keep the status quo in that regard. A few other call sites may have assumed zero prices that will now receive an error, but that is _probably_ a good thing.

### DIFF
--- a/pallets/cash/src/internal/liquidity.rs
+++ b/pallets/cash/src/internal/liquidity.rs
@@ -250,7 +250,6 @@ mod tests {
 
     #[test]
     fn test_has_liquidity_to_reduce_asset_mismatch() {
-        // TODO: This should fail -- and currently doesn't
         new_test_ext().execute_with(|| {
             assert_eq!(
                 has_liquidity_to_reduce_asset::<Test>(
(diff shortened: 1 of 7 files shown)


pallets/cash/src/internal/liquidity.rs
```diff
@@ -239,11 +258,34 @@ mod tests {
...
                 ),
-                Ok(true)
+                Err(Reason::AssetQuantityMismatch)
+            );
+        })
+    }
+
+    #[test]
+    fn test_has_liquidity_to_reduce_asset_unpriced() {
+        // TODO: This should fail -- and currently doesn't
```
jflatow: Still?

hayesgm: Thanks, removed.

jflatow (approved):

PR #287, opened 2021-04-20
Event processing should handle reorgs / undo

jflatow:
This still needs more tests and to implement undo lock functionality. It especially needs to be tested that events are applied correctly after ingestion. However, it now compiles / passes existing tests, and could be merged to reduce future conflicts.


hayesgm (approved):
lgtm. awesome improvements.

PR #288, opened 2021-04-21
Cash Pipeline

hayesgm:
This patch adds `CashPipeline` which lets us string together multiple operations (e.g. transfer asset, transfer cash) in a non-effectful way until we're ready to commit. This allows us to check conditions on complex interactions (e.g. the liquidity after a liquidation) without fear of changing state to do so. This, overall, significantly reduces our total LoC.


pallets/cash/src/internal/balance_helpers.rs
```diff
@@ -0,0 +1,535 @@
+use crate::{
+    reason::{MathError, Reason},
+    types::{
+        AssetAmount, AssetBalance, AssetQuantity, CashPrincipal, CashPrincipalAmount, Quantity,
+    },
+};
+use our_std::{cmp::min, convert::TryFrom};
+
+// XXX use Balances instead of raw balances everywhere and put all fns on types?
```
jflatow, line 9: I removed this in my branch, I think we're better off with these functions as they make the intentions very clear - I started to convert it to using impl fns on the types and it became a lot less obvious

pallets/cash/src/internal/liquidate.rs
```diff
@@ -0,0 +1,553 @@
...
+    let seize_amount = amount
+        .mul_factor(liquidation_incentive)?
+        .mul_price(get_price::<T>(asset.units())?)?
+        .div_price(
+            get_price::<T>(collateral_asset.units())?,
+            collateral_asset.units(),
+        )?;
+
+    require!(
+        !liquidity::has_non_negative_liquidity::<T>(borrower)?,
```
jflatow: Probably worth just adding a `has_negative_liquidity` wrapper? I think require double negative is harder to parse

pallets/cash/src/internal/liquidate.rs
```diff
@@ -0,0 +1,553 @@
...
+        Reason::SufficientLiquidity
+    );
+
+    Effectful::new()
+        .transfer_asset::<T>(liquidator, borrower, asset.asset, amount)?
+        .transfer_asset::<T>(borrower, liquidator, collateral_asset.asset, seize_amount)?
+        .check_liquidity::<T>(liquidator)?
```
jflatow: I feel like this should be `has_non_negative_liquidity` as well? `check_liquidity` is a bit ambiguous

jflatow (approved):
Looking great, glad you mustered the strength to do this! 💪

pallets/cash/src/internal/liquidate.rs
hayesgm: I might rework these once we get lock and extract moved over, since they can mostly all go away at that point.

pallets/cash/src/internal/liquidate.rs
hayesgm: I'm just wondering what the misinterpretation could be here? That we could calculate it? Without the work `check` it becomes less clear, IMO, but with more, it gets pretty verbose, e.g. `check_account_has_sufficient_liquidity`

PR #289, opened 2021-04-21
add more data about cash to the cashdata rpc

coburncoburn:


runtime/src/lib.rs
```diff
@@ -255,7 +252,7 @@ parameter_types! {
 }
 
 parameter_types! {
-    pub const TransactionByteFee: Balance = 10_000_000; // XXX how are we doing weights/fees for substrate?
+    pub const TransactionByteFee: u128 = 10_000_000; // XXX how are we doing weights/fees for substrate?
```
coburncoburn: is this going to explode something? Balance was just a type alias for u128 here, and it conflicted with importing our balance type

jflatow: 🆒 if we need this to compile can you just delete the `XXX` and maybe instead `note that unused but we need this in order to compile`?

jflatow (approved):

PR #291, opened 2021-04-21
Limit Liqudations by Spec

hayesgm:
This patch fixes liqudiations by limiting when they can occur, specifically prohibiting you from liquidating a borrow into a positive position (close > 100%) and preventing you from liquidating collateral into a borrow position (seize > 100% collateral). These are important fixes for real-life.

We will add better unit tests, but we are waiting on fixes for scenarios to test the current scenarios for this behavior.

### DIFF
--- a/pallets/cash/src/core.rs
+++ b/pallets/cash/src/core.rs
@@ -57,6 +57,17 @@ macro_rules! require {
 }
 
 #[macro_export]
+macro_rules! must {
+    ($expr:expr, $reason:expr) => {
+        if !$expr {
+            core::result::Result::Err($reason)
+        } else {
+            core::result::Result::Ok(())
+        }
+    };
+}
+
+#[macro_export]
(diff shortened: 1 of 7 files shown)


coburncoburn (approved):
approve!

jflatow (approved):
Awesome job 🥇

pallets/cash/src/core.rs
```diff
@@ -56,6 +56,17 @@ macro_rules! require {
     };
 }
 
+#[macro_export]
+macro_rules! must {
```
maxwolff, line 60: same as require but "must" is easier to think?

hayesgm: Must returns a `Result<(), Reason>` where `require!` is a unit value itself `if { ... };`. In some other code, I really want a `Result<(), Reason>` but I don't want to change the `require!` macro that we use in a lot of places.

PR #293, opened 2021-04-22
ninja edit to make new enum compatible with existing testnet storage

coburncoburn:
put Dot ( never been used ) where Gate ( also never used ) was so Eth ( constantly used ) remains in second position of the enum.


jflatow (commented):
Clever 😸 Good option, though I think we should probably add chains in order of Starport added, and not hacked around for this... I also think we ought to understand/practice how to do one of these migrations https://hackmd.io/BQt-gvEdT66Kbw0j5ySlWw e.g. https://github.com/hicommonwealth/substrate/pull/14/files 🤷

hayesgm:
FWIW, I wrote a `on_runtime_upgrade` fn, and I couldn't get it to run in scenarios. Not sure why but they are easy to do conceptually.

coburncoburn:
this is insufficient to solve the problem

PR #294, opened 2021-04-22
Use last block property from runtime interface as default

jflatow:


hayesgm (approved):
lgtm. Can you updates `types.json`?

PR #302, opened 2021-04-23
add configurable fetch deadline for eth client

toninorair:
Fixing one of the  XXX comments and removing the hard-coded deadline - 2s for fetching eth data. Set it to 10s by default.


pallets/runtime-interfaces/src/lib.rs
```diff
@@ -25,6 +25,7 @@ pub struct ValidatorConfig {
 impl Config {
     pub fn update(&mut self, new: Config) {
         self.eth_starport_address = new.eth_starport_address;
+        self.eth_starport_parent_block = new.eth_starport_parent_block;
```
toninorair: I suspect this thing is missing, not related to this exact pr though.

jflatow: Already fixed on `develop`

jflatow (approved):

ethereum-client/src/lib.rs
```diff
@@ -243,15 +245,19 @@ pub fn get_block(
...
     let params = vec![block_num.into(), true.into()];
-    let response_str: String = send_rpc(server, "eth_getBlockByNumber".into(), params)?;
+    let response_str: String = send_rpc(server, "eth_getBlockByNumber".into(), params, deadline)?;
     let response = deserialize_get_block_by_number_response(&response_str)?;
     response.result.ok_or(EthereumClientError::JsonParseError)
 }
 
-pub fn get_latest_block_number(server: &str) -> Result<u64, EthereumClientError> {
-    let response_str: String = send_rpc(server, "eth_blockNumber".into(), vec![])?;
+pub fn get_latest_block_number(server: &str, deadline: u64) -> Result<u64, EthereumClientError> {
```
jflatow, line 269: Oh I don't think we want to have to pass this in?

toninorair: @jflatow do you mean that the deadline should be read from runtime_interfaces inside ethereum-client?

jflatow: I think it should just be a compile time `const`?

toninorair: Ok, so the overthinking here, I will move it to just a const var

PR #303, opened 2021-04-23
Resolve some open questions / standing issues in rates

jflatow:


pallets/cash/src/rates.rs
```diff
@@ -44,27 +42,22 @@ impl APR {
...
+    pub fn compound(self, dt: Timestamp) -> Result<Factor, MathError> {
+        let one = Factor::ONE;
+        let x_1 = self.simple(dt)?;
+        let x_2 = x_1.mul_uint(x_1.0);
+        let x_3 = x_1.mul_uint(x_1.0).mul_uint(x_1.0);
+        Ok(Factor(
+            BigUint::from_uint(one.0)
+                .add(x_1.mul_uint(1))
+                .add(x_2.div_uint(2)?.div_uint(one.0)?)
+                .add(x_3.div_uint(6)?.div_uint(one.0)?.div_uint(one.0)?)
```
coburncoburn, line 55: again, dividing by identity?

pallets/cash/src/rates.rs
```diff
@@ -44,27 +42,22 @@ impl APR {
...
-        }
+    /// Official approximation of e^(r*dt) using e^x = 1 + x + x^2/2 + x^3/6 + ...
+    pub fn compound(self, dt: Timestamp) -> Result<Factor, MathError> {
+        let one = Factor::ONE;
+        let x_1 = self.simple(dt)?;
+        let x_2 = x_1.mul_uint(x_1.0);
+        let x_3 = x_1.mul_uint(x_1.0).mul_uint(x_1.0);
+        Ok(Factor(
+            BigUint::from_uint(one.0)
+                .add(x_1.mul_uint(1))
```
coburncoburn, line 53: I guess multiplying by one returns a different type, and it's the one you want? ( why multiply by identity? )

pallets/cash/src/rates.rs
```diff
@@ -44,27 +42,22 @@ impl APR {
...
-        let t4 = scaled_rate * scaled_rate * scaled_rate / 6; // x^3 / 6
-        let unscaled = t1 + t2 + t3 + t4;
-        let scaled: BigInt = unscaled / index_scale / index_scale;
-        if let Some(raw) = scaled.to_u128() {
-            Ok(CashIndex(raw))
-        } else {
-            Err(MathError::Overflow)
-        }
+    /// Official approximation of e^(r*dt) using e^x = 1 + x + x^2/2 + x^3/6 + ...
+    pub fn compound(self, dt: Timestamp) -> Result<Factor, MathError> {
```
coburncoburn, line 46: I think this was subject to a lot of discussions that I didn't really participate in so maybe I am missing some context, but LGTM

coburncoburn (approved):
LGTM, left some comments on exp math that probably is probably right, but is a bit surprising it's not more ergonomic by now  since I thought we had a library for it.

pallets/cash/src/rates.rs
jflatow: It's just this one is a weird numerical calculation - you have 4 terms with different numbers of factors - I think this is actually the clearest way to write it...at least the best I found

pallets/cash/src/rates.rs
jflatow: Yeah its just to avoid adding another function variant to accept the Factor (tricky but x_1 is a `Factor`, x_2 and x_3 are `factor::BigUint`s)

pallets/cash/src/rates.rs
jflatow: I think that was probably mostly around how to do the approximation...this is just updating it to use the math functions that were added later and return a `Factor`

hayesgm (approved):
lgtm

PR #313, opened 2021-04-27
Fix m8 Failure

hayesgm:
Fix m8 Failure
    
This patch fixes failures running m8 due to the enum shifting between versions. We add a reserved filler into the enums that can later be set to a different value, e.g. for another chain. We try to fill all of the slots since we would prefer that the numbering of chains stays constant and doesn't shift (e.g. it's #0 for ChainAccounts but #1 for ChainAssets and #5 for ChainBlocks).

## Previous Notes

Currently we cannot run m7 wasm in the native shell. This is likely due to some incompatability in the host functions versus the previously built wasm code. This new test shows off that behavior (it was broken with a Wasmi Trap). We fix the issue by changing the Dot/Eth order in chains.rs (not sure why this is the fix, but it worked). Note: this test now traps on missing a host function for getting cash data, but this is **not of any concern**. The original error was that blocks could not form, and we have a partial fix here that warrants more follow up. That is: why did this fix work? How can we make it work without changing the enum orders?

Again, the current error in the test is not the issue. The issue is "fixed" as we can mine blocks, but we need to figure out why this fix worked and if we can fix it without the "hack".

### DIFF
--- a/pallets/cash/src/chains.rs
+++ b/pallets/cash/src/chains.rs
@@ -17,6 +17,7 @@ use types_derive::{type_alias, Types};
 #[derive(Serialize, Deserialize)] // used in config
 #[derive(Copy, Clone, Eq, PartialEq, PartialOrd, Ord, Encode, Decode, RuntimeDebug, Types)]
 pub enum ChainId {
+    Reserved,
     Eth,
     Dot,
 }
@@ -24,6 +25,7 @@ pub enum ChainId {
 impl ChainId {
     pub fn to_account(self, addr: &str) -> Result<ChainAccount, Reason> {
         match self {
+            ChainId::Reserved => Err(Reason::Unreachable),
             ChainId::Eth => Ok(ChainAccount::Eth(Ethereum::str_to_address(addr)?)),
             ChainId::Dot => Ok(ChainAccount::Dot(Polkadot::str_to_address(addr)?)),
         }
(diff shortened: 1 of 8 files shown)


coburncoburn:
One problem is this: integration tests build their own chainspec, so the chainspec puts validator keys into genesis config using the new enums, but the wasm expects the old ones.

pallets/cash/src/chains.rs
```diff
@@ -73,12 +80,13 @@ impl ChainId {
             ChainId::Eth => ChainBlock::Eth(
                 runtime_interfaces::config_interface::get_eth_starport_parent_block(),
             ),
-            _ => panic!("XXX not implemented"),
+            ChainId::Reserved | ChainId::Dot => panic!("xxx not supported"),
```
coburncoburn, line 83: may save some time debugging later if the panic has the faulty chain it

coburncoburn (approved):
👍

coburncoburn:
really does beg the question of how these tests will hold up if the code to generate a compatible chainspec is not available. Maybe there is a way to use wasm to build the chainspec

jflatow (approved):

pallets/cash/src/chains.rs
hayesgm: yeah. I also want to mostly remove panics, but some of these are hard to inhibit without having a concrete value to put in

PR #315, opened 2021-04-28
Miner cumulative, asset modified events, and some other fixes

jflatow:


pallets/cash/src/internal/set_yield_next.rs
```diff
@@ -15,48 +15,38 @@ use types_derive::Types;
...
 pub fn set_yield_next<T: Config>(
     next_yield: APR,
     next_yield_start: Timestamp,
 ) -> Result<(), Reason> {
-    let now = get_now::<T>();
-    let change_in_time = next_yield_start
-        .checked_sub(now)
-        .ok_or(Reason::TimeTravelNotAllowed)?;
+    let now = get_recent_timestamp::<T>()?;
+    let min_t = now + MIN_NEXT_SYNC_TIME;
```
hayesgm: We're okay with the unchecked addition since it's a) just a check, and b) the paramis a const?

pallets/cash/src/internal/set_yield_next.rs
```diff
@@ -15,48 +15,38 @@ use types_derive::Types;
...
-    // Read Require NextAPRStartNow+ParamsMinNextSyncTime
-    require!(
-        next_yield_start >= now + MIN_NEXT_SYNC_TIME,
-        Reason::SetYieldNextError(SetYieldNextError::TimestampTooSoonToNow)
-    );
-
-    // XXX fix this up
-    // Read NextYieldIndex=GetCashYieldIndexAt(NextAPRStart)
-    let next_yield_index: CashIndex = get_cash_yield_index_after::<T>(change_in_time)?;
+    let dt = next_yield_start - now;
```
hayesgm: Same with the unchecked sub.

pallets/cash/src/reason.rs
```diff
@@ -60,6 +58,7 @@ pub enum Reason {
     SetYieldNextError(SetYieldNextError),
     SignatureAccountMismatch,
     SignatureMismatch,
+    TimestampMissing,
```
hayesgm, line 61: I think we should only append to the end. Shuffling enums should be a no-go

hayesgm (commented):
Looks awesome. Few minor comments.

pallets/cash/src/lib.rs
```diff
@@ -207,6 +207,12 @@ decl_storage! {
         /// Miner of the current block.
         Miner get(fn miner): Option<ChainAccount>;
 
+        /// Mapping of total principal paid to each miner.
+        MinerCumulative get(fn miner_cumulative): map hasher(blake2_128_concat) ChainAccount => CashPrincipalAmount;
+
+        /// Total principal paid to all miners.
+        TotalMinerCumulative get(fn total_miner_cumulative): CashPrincipalAmount;
```
coburncoburn: would it be better to iterate MinerCumulative in rpc's and then not need this storage?

pallets/cash/src/internal/set_yield_next.rs
jflatow: Yeah I was thinking about that too, I didn't change it though. I guess the expectation is that current timestamp + MIN_NEXT_SYNC_TIME must fit into a u64?

pallets/cash/src/internal/set_yield_next.rs
jflatow: This one is safe I think since we check `next_yield_start >= now + C`  where `C > 0` above

pallets/cash/src/lib.rs
jflatow: That's probably excessively inefficient? I mean no worse the liquidity endpoint, but we should try to minimize/eliminate such endpoints I think...I think real answer longer term is events and apis (subscan, subquery should actually be able to handle this with proper eventing)

pallets/cash/src/reason.rs
jflatow: The `Reason` type isn't stored anywhere so it shouldn't break an upgrade (I think in general we should/could promise that about Reason now, although it did *used* to be in the Failed processing event state). I also think there's some benefit to grouping these and still dream of the day we might generate these codes. But I guess the argument is that it breaks (or could break) the client

pallets/cash/src/lib.rs
coburncoburn: it would be like... 13 reads in a dedicated rpc node, a single trx request probably has more than that many reads in it.

coburncoburn: also we'll be wanting each miner's cumulative to be up to date, so those per-validator reads will be happening anyways

coburncoburn: or we could just write TotalMinerCumulative anyways 🤷

pallets/cash/src/internal/set_yield_next.rs
hayesgm: Though, with a potential wrap around.... I wonder based on every other inefficiency and the fact this already returns a `Result` if we should just do the check

pallets/cash/src/lib.rs
hayesgm: Is this supposed to be removed?

jflatow: Yep 🎣

pallets/cash/src/internal/set_yield_next.rs
```diff
@@ -15,48 +15,42 @@ use types_derive::Types;
...
-        next_yield_start >= now + MIN_NEXT_SYNC_TIME,
-        Reason::SetYieldNextError(SetYieldNextError::TimestampTooSoonToNow)
-    );
-
-    // XXX fix this up
-    // Read NextYieldIndex=GetCashYieldIndexAt(NextAPRStart)
-    let next_yield_index: CashIndex = get_cash_yield_index_after::<T>(change_in_time)?;
+    let dt = next_yield_start
+        .checked_sub(now)
+        .ok_or(MathError::Underflow)?;
```
hayesgm, line 52: FWIW, this could probably be an `expect` due to your reasoning from before.

pallets/cash/src/reason.rs
hayesgm: Though we've seen clients break~ish when they've used outdated reasons, which is harder that the reason comes up wrong/broken. But I could see this being a "safe to modify" enum if we wanted to go that way.

hayesgm (approved):
lgtm

PR #316, opened 2021-04-28
Fix substrate tests ?

toninorair:
An attempt to fix parallelization error in a logical way

### DIFF
--- a/pallets/cash/src/internal/assets.rs
+++ b/pallets/cash/src/internal/assets.rs
@@ -263,7 +263,10 @@ mod tests {
                 },
             ];
 
-            assert_eq!(super::get_assets::<Test>()?, assets);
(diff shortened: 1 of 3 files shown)


jflatow (approved):
Very nice 💪

jflatow (commented):
I took your get_assets changed and it was merged with my branch already. I also changed CI to run single threaded tests for now, should we keep these other fixes or now think they were a red herring?

toninorair:
> I took your get_assets changed and it was merged with my branch already. I also changed CI to run single threaded tests for now, should we keep these other fixes or now think they were a red herring?

Those fixes are improvements in any case, it doesn't hurt to have them. The real cause is `a_receive_chain_blocks` and `b_receive_chain_blocks` changing env var just to get signatures

PR #319, opened 2021-04-29
Fix tests parallelism

toninorair:
The real fix for tests parallelism 
instead of changing validator env vars just to get signature in a convenient way, this approach gets signature directly.

### DIFF
--- a/pallets/cash/src/tests/mod.rs
+++ b/pallets/cash/src/tests/mod.rs
@@ -1,15 +1,19 @@
 #![allow(non_upper_case_globals)]
 
 pub use codec::{Decode, Encode};
-use ethereum_client::encode_block_hex;
 pub use frame_support::{assert_err, assert_ok, dispatch::DispatchError};
 pub use hex_literal::hex;
-pub use our_std::convert::TryInto;
-pub use our_std::{iter::FromIterator, str::FromStr};
-use parking_lot::RwLock;
+pub use our_std::{
+    sync::Arc,
+    convert::TryInto,
+    iter::FromIterator, str::FromStr,
+};
 pub use sp_core::crypto::AccountId32;
+use parking_lot::RwLock;
+use secp256k1::SecretKey;
 use sp_core::offchain::testing;
-use std::sync::Arc;
+
+use ethereum_client::encode_block_hex;
 
 pub mod assets;
 pub mod common;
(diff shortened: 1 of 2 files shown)


pallets/cash/src/tests/mod.rs
```diff
@@ -159,11 +160,12 @@ pub fn initialize_storage() {
...
-        "ETH_KEY",
-        "6bc5ea78f041146e38233f5bc29c703c1cec8eaaa2214353ee8adf7fc598f23d",
-    );
-    validator_sign::<Test>(data)
+    let private_key =
+        hex::decode("6bc5ea78f041146e38233f5bc29c703c1cec8eaaa2214353ee8adf7fc598f23d").unwrap();
+    let secret_key = SecretKey::parse_slice(&private_key).unwrap();
+    let signature_raw = gateway_crypto::eth_sign(data, &secret_key, false);
+    let signature = ChainSignature::Eth(signature_raw);
+    Ok(signature)
```
jflatow, line 177: Hm this doesn't exercise `validator_sign` anymore though, but maybe that's ok if it's tested elsewhere?

toninorair: It returns signature, I think that's actually the right way to do it. To get a validator signature, there is no need to change env var for the convenience of calling `validator_sign`.  That was casing tests failures unless they all run serially, now I see all tests passing

toninorair: @jflatow do you think there is a need to change env var?

jflatow: I think that makes sense, its just like how do we test `validator_sign`?

toninorair: should it be tested at unit level? it realistically just the wrapper around sign_message --> eth_sign. It's like if we want to run tests in parallel then env vars should not be changed. 
Alternatively tests that change vars can be marked with some special attribute and run separately. But the whole test suite, there is only one such test now.

toninorair:
@jflatow let's merge it in? It is an improvement anyway and allows to run tests in parallel

jflatow:
I don't think this fixes the issue, closing

toninorair:
> I don't think this fixes the issue, closing

but it did, it was actually fixing the issue with running tests in parallel by getting signature without changing the env var

toninorair:
It's out of sync now, I will clean it up.

jflatow:
OK I rebased and can confirm this does seem to work now, but doesn't exercise the `validator_sign` path, so let's fix again as we try to get rid of the env var completely next

jflatow:
Sorry 🎠, this got closed when I deleted the base moon branch and now I can't reopen, opening another PR

PR #321, opened 2021-04-29
Upgrade to m9 Scenario

hayesgm:
This patch gets the upgrade tests working. They are pretty crazy due to the stateful API connections that make it quite difficult to send an RPC that changes the encoding/decoding scheme, since they can't really be on version 1 (since they crash after changeover as soon as they see an event) nor version 2 (since they crash trying to init with an old types.json). You can try to blend the types.json from both versions (we do this) but you'll still get truly conflicting decoding schemes. It's just a mess so we actually mostly try to disconnect the APIs and reconnect them after the change. It's a bit flakey but it works somewhat okay.

With this strategy, we fix the previous upgrade scenarios and we add an m8 -> m9 scenario.

We were unable to get into the state seen on test-net (not sure how to repro), and thus we still don't have full confidence this upgrade will fix the issue, but we see no reason that it should specifically break it, either.

### DIFF
--- a/integration/__tests__/gov_scen.js
+++ b/integration/__tests__/gov_scen.js
@@ -25,7 +25,6 @@ buildScenarios('Gov Scenarios', gov_scen_info, [
     }
   },
   {
-    skip: true,
     name: "Upgrade Chain WASM [Allow Next Code]",
     info: {
       versions: ['m7'],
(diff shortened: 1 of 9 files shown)


jflatow (approved):
Great work 👨‍🍳 🛳️

PR #325, opened 2021-05-03
Heuristic Reorg Benchmarks 

maxwolff:
many things impact the gas consumption of a reorg, I take an estimate for the sake of dev time. 

* if the reorg has enough support to get applied, I just multiply num_reorg_events * avg_gas_cost (ignore # blocks, difference btwn forward & backward cost). I didn't write an actual benchmark bc figuring out the validators btwn test env and benchmark env was getting annoying 
* otherwise, I use an actual benchmark for adding a pending reorg

I think we should check if there is sufficient support for estimating the for the receive_chain_blocks gas as well.


pallets/cash/src/internal/events.rs
```diff
@@ -573,13 +568,20 @@ mod tests {
...
+                    hash: latest_hash,
                     parent_hash: [4; 32],
                     number: 5,
                     events: vec![],
                 },
             ]);
 
+            let reorg = ChainReorg::Eth {
+                from_hash: latest_hash,
+                to_hash: real_block_hash,
```
maxwolff, line 698: just revise test for clarity, was accidentally using [5;32] for both here

PR #329, opened 2021-05-17
Const compile config

jflatow:
This is on top of https://github.com/compound-finance/gateway/pull/326 squashed.

We may want to separate into `params/{mod,mainnet,testnet}.rs` files and we can bike shed about whether to always define the `params::{mainnet, testnet}` modules or only when the corresponding feature flag is triggered and the values are actually used.


jflatow:
Superceded by the 'runtime with 2-step upgrade' approach in #335

PR #330, opened 2021-05-18
Polygon support

waynenilsen:
Changes to support Polygon/MATIC

Open questions

We need to really figure out what we want to do about finality on Polygon. Based on our conversation with them it sounds like we just take the same approach as we do with Ethereum but that needs to be made into a firm commitment. We are on the hook for sharing our starport specification with them.

Governance of the starport and cash token remain mysterious on Polygon. Eventually of course we want to move this onto  Gateway but the path there is currently somewhat unclear. Perhaps via notices. Needs to be sketched out in detail.

### DIFF
--- a/ethereum-client/src/events.rs
+++ b/ethereum-client/src/events.rs
@@ -312,6 +312,7 @@ pub fn decode_event(topics: Vec<String>, data: String) -> Result<EthereumEvent, 
         .iter()
         .map(|topic| decode_topic(topic).ok_or(EventError::InvalidTopic))
         .collect::<Result<Vec<ethabi::Hash>, _>>()?;
+
     match topic_hashes.first().ok_or(EventError::InvalidTopic)? {
         t if *t == *LOCK_EVENT_TOPIC => {
             let log: ethabi::Log = LOCK_EVENT
--- a/ethereum/contracts/Starport.sol
+++ b/ethereum/contracts/Starport.sol
@@ -14,8 +14,8 @@ contract Starport {
 
     address immutable public admin;
     address constant public ETH_ADDRESS = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
-    bytes4 constant MAGIC_HEADER = "ETH:";
-    string constant ETH_CHAIN = "ETH";
+    bytes4 immutable chainIdHeader; // = "ETH:" on Ethereum
+    bytes32 immutable chainId; // = "ETH" on Ethereum
     address[] public authorities;
     mapping(address => uint) public supplyCaps;
 
@@ -35,9 +35,23 @@ contract Starport {
     event ExecuteProposal(string title, bytes[] extrinsics);
     event NewSupplyCap(address indexed asset, uint supplyCap);
 
-    constructor(ICash cash_, address admin_) {
+    constructor(ICash cash_, address admin_, bytes32 chainId_, bytes4 chainIdHeader_) {
         cash = cash_;
         admin = admin_;
+        chainId = chainId_;
+        chainIdHeader = chainIdHeader_;
+    }
+
+    function bytes32ToString(bytes32 _bytes32) public pure returns (string memory) {
+        uint8 i = 0;
+        while (i < 32 && _bytes32[i] != 0) {
+            i++;
+        }
+        bytes memory bytesArray = new bytes(i);
+        for (i = 0; i < 32 && _bytes32[i] != 0; i++) {
+            bytesArray[i] = _bytes32[i];
+        }
+        return string(bytesArray);
     }
 
     /**
@@ -51,7 +65,7 @@ contract Starport {
      * @param asset The asset to lock in the Starport
      */
     function lock(uint amount, address asset) external {
-        lockTo(amount, asset, ETH_CHAIN, toBytes32(msg.sender));
+        lockTo(amount, asset, bytes32ToString(chainId), toBytes32(msg.sender));
     }
 
     /**
@@ -77,7 +91,7 @@ contract Starport {
      * @dev Use `lock` to lock CASH or collateral assets.
      */
     function lockEth() public payable {
-        lockEthTo(ETH_CHAIN, toBytes32(msg.sender));
+        lockEthTo(bytes32ToString(chainId), toBytes32(msg.sender));
     }
 
     /*
@@ -234,10 +248,10 @@ contract Starport {
         isNoticeInvoked[noticeHash] = true;
 
         require(notice.length >= 100, "Must have full header"); // 4 + 3 * 32
-        require(notice[0] == MAGIC_HEADER[0], "Invalid header[0]");
-        require(notice[1] == MAGIC_HEADER[1], "Invalid header[1]");
-        require(notice[2] == MAGIC_HEADER[2], "Invalid header[2]");
-        require(notice[3] == MAGIC_HEADER[3], "Invalid header[3]");
+        require(notice[0] == chainIdHeader[0], "Invalid header[0]");
+        require(notice[1] == chainIdHeader[1], "Invalid header[1]");
+        require(notice[2] == chainIdHeader[2], "Invalid header[2]");
+        require(notice[3] == chainIdHeader[3], "Invalid header[3]");
 
         (uint noticeEraId, uint noticeEraIndex, bytes32 noticeParent) =
             abi.decode(notice[4:100], (uint, uint, bytes32));
--- a/integration/__tests__/lock_scen.js
+++ b/integration/__tests__/lock_scen.js
@@ -6,7 +6,8 @@ let lock_scen_info = {
   tokens: [
     { token: 'usdc', balances: { ashley: 1000 } }
   ],
-  validators: ['alice', 'bob']
+  validators: ['alice', 'bob'],
+  native: true
 };
 
 async function getCash({ ashley, usdc, cash, chain, starport }) {
--- a/integration/__tests__/matic_scen.js
+++ b/integration/__tests__/matic_scen.js
@@ -1,31 +1,118 @@
 const { buildScenarios } = require('../util/scenario');
 const { getNotice } = require('../util/substrate');
 
+let now = Date.now();
+
 let lock_scen_info = {
     tokens: [
         { token: 'usdc', balances: { ashley: 1000 } },
-        { token: 'maticZrx', balances: { darlene: 1000 } }
+        { token: 'maticZrx', balances: { darlene: 1000000 } },
+        { token: 'zrx', balances: { bert: 1000000 } },
+        { token: 'comp' }
     ],
     validators: ['alice', 'bob'],
-    actors: ['ashley', 'darlene'],
+    actors: ['ashley', 'bert', 'chuck', 'darlene', 'edward'],
     chain_opts: {
         matic: {
             name: 'matic',
-            provider: 'ganache', // [env=PROVIDER]
+            provider: 'ganache',
             ganache: {
                 opts: {},
                 web3_port: null
             },
         },
     },
+    native: true, // for log messages - removes wasm:stripped - no need for wasm no forkless upgrade testing here
+    freeze_time: now,
+    initial_yield: 300,
+    initial_yield_start_ms: now
 };
 
 buildScenarios('Matic', lock_scen_info, [
     {
-        skip: true, // TODO FIX SCEN
-        name: 'Matic',
-        scenario: async ({ darlene, maticZrx }) => {
-            await darlene.lock(1000, maticZrx);
+        name: 'Matic Lock',
+        scenario: async ({ashley, usdc, darlene, maticZrx }) => {
+            // Lock on Polygon
+            let balance, liquidity;
+            await darlene.lock(100, maticZrx);
+            balance = await darlene.balanceForToken(maticZrx);
+            let darleneLiquidity = await darlene.liquidity();
+            expect(balance).toEqual(100);
+            expect(darleneLiquidity).toBeGreaterThan(0);
+
+            // lock on ethereum
+            await ashley.lock(1000, usdc);
+            balance = await ashley.balanceForToken(usdc);
+            liquidity = await ashley.liquidity();
+            expect(balance).toEqual(1000);
+            expect(liquidity).toBeGreaterThan(0);
+
+            // lock on polygon again
+            await darlene.lock(100, maticZrx);
+            balance = await darlene.balanceForToken(maticZrx);
+            liquidity = await darlene.liquidity();
+            expect(balance).toEqual(200); // double the asset balance (no interest, 0 utilization)
+            expect(liquidity).toEqual(2*darleneLiquidity); // liquidity must have gone up
+        }
+    },
+    {
+        name: 'Collateral Borrowed Interest Lump Sum',
+        scenario: async ({ darlene, chuck, cash, chain, usdc, maticZrx}) => {
+            // await prices.postPrices();
+            let balance, liquidity;
+            await chain.setFixedRate(usdc, 500); // 5% APY fixed
+            await darlene.lock(10000, maticZrx);
+            balance = await darlene.balanceForToken(maticZrx);
+            let liquidity1 = await darlene.liquidity();
+            expect(balance).toEqual(10000);
+            expect(liquidity1).toBeGreaterThan(0);
+            // now we know everything is in order, make the transfer, let's say half our liquidity worth
+            await darlene.transfer(1000, usdc, chuck);
+            expect(await darlene.chainBalance(usdc)).toEqual(-1000);
+            expect(await chuck.chainBalance(usdc)).toEqual(1000);
+            expect(await darlene.chainBalance(cash)).toBeCloseTo(-0.01, 2); // 1¢ transfer fee
+            expect(await chuck.chainBalance(cash)).toEqual(0);
+            await chain.accelerateTime({years: 1});
+            expect(await darlene.chainBalance(usdc)).toEqual(-1000);
+            expect(await chuck.chainBalance(usdc)).toEqual(1000);
+            expect(await darlene.chainBalance(cash)).toBeCloseTo(-51.53272669767585, 3); // -50 * Math.exp(0.03) - 0.01
+            expect(await chuck.chainBalance(cash)).toBeCloseTo(51.52272669767585, 3); // 50 * Math.exp(0.03)
+        }
+    },
+    {
+        name: "Extract Collateral",
+        scenario: async ({ darlene, maticZrx, chain, maticStarport }) => {
+            let assetInfo = await maticZrx.getAssetInfo();
+
+            let balance, liquidity;
+            // start with 1,000,000
+            // lock 10,000
+            // on matic = 990,000
+            // on gateway = 10,000
+            await darlene.lock(10000, maticZrx);
+            balance = await darlene.balanceForToken(maticZrx);
+            expect(balance).toEqual(10000); // gateway balance not yet debited
+            expect(await darlene.tokenBalance(maticZrx)).toEqual(990000); // matic balance not yet credited
+            liquidity = await darlene.liquidity();
+            expect(liquidity).toBeGreaterThan(0);
+
+            // request extract 50 maticZrx
+            let extract = await darlene.extract(50, maticZrx);
+            let notice = getNotice(extract);
+            let signatures = await chain.getNoticeSignatures(notice);
+
+            // before we submit - query the balances to ensure they are as expected
+            balance = await darlene.balanceForToken(maticZrx);
+            expect(balance).toEqual(9950); // gateway balance debited
+            expect(await darlene.tokenBalance(maticZrx)).toEqual(990000); // matic balance not yet credited
+            // now invoke the notice!
+            await maticStarport.invoke(notice, signatures);
+            // balances are updated everywhere, should now be
+            // on matic = 990,050
+            // on gateway = 9,950
+            balance = await darlene.balanceForToken(maticZrx);
+            expect(balance).toEqual(9950); // gateway balance debited
+            expect(await darlene.tokenBalance(maticZrx)).toEqual(990050); // matic balance now credited
         }
     },
 ]);
--- a/integration/util/scenario/actor.js
+++ b/integration/util/scenario/actor.js
@@ -16,7 +16,7 @@ class Actor {
   }
 
   chain() {
-    return this.ctx.chains.find(this.chainName());
+    return this.ctx.chains.find(this.chainName);
   }
 
   show() {
@@ -32,11 +32,13 @@ class Actor {
   }
 
   toChainAccount() {
-    return { Eth: this.ethAddress() };
+    const returnValue = {};
+    returnValue[this.chain().nameAsPascalCase()] = this.ethAddress();
+    return returnValue;
   }
 
   toTrxArg() {
-    return `Eth:${this.ethAddress()}`;
+    return `${this.chain().nameAsPascalCase()}:${this.ethAddress()}`;
   }
 
   declareInfo() {
@@ -59,14 +61,16 @@ class Actor {
   }
 
   async sign(data) {
-    return await this.ctx.eth.sign(data, this);
+    return await this.chain().sign(data, this);
   }
 
   async signWithNonce(data) {
     let currentNonce = await this.nonce();
     let signature = await this.sign(`${currentNonce}:${data}`);
+    const signatureData = {};
+    signatureData[this.chain().nameAsPascalCase()] = [this.ethAddress(), signature];
 
-    return [ { Eth: [ this.ethAddress(), signature ] }, currentNonce ];
+    return [ signatureData, currentNonce ];
   }
 
   async runTrxRequest(trxReq) {
@@ -77,7 +81,7 @@ class Actor {
   }
 
   async ethBalance() {
-    return await this.ctx.eth.ethBalance(this);
+    return await this.chain().ethBalance(this);
   }
 
   async tokenBalance(tokenLookup) {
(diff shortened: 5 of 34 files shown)


integration/util/scenario/chain_spec.js
```diff
@@ -34,8 +34,10 @@ async function baseChainSpec(validatorsInfoHash, tokensInfoHash, ctx) {
       eth_address: v.eth_account
   }));
 
-  let assets = tokens.all().map((token) => ({
-    asset: `Eth:${token.ethAddress()}`,
+  let assets = tokens.all()
+      .filter(token => token.symbol.toUpperCase() !== 'CASH')
```
waynenilsen, line 68: dev note - i am 99% sure we want this filter but before we had CASH in `SupportedAssets` storage

resulted in having to do some downstream fixes

node/src/command.rs
```diff
@@ -136,6 +136,7 @@ pub fn run() -> sc_cli::Result<()> {
             runtime_interfaces::initialize_validator_config(
                 cli.gateway.eth_key_id.clone(),
                 cli.gateway.eth_rpc_url.clone(),
+                cli.gateway.matic_rpc_url.clone(), // XXX todo: do we even actually need new keys for matic/polygon? I think eth_key_id should be fine / enough.
```
toninorair: we probably need it, if infura supports matic as promised, it will be a separate URL for sure.

integration/util/scenario/chains.js
```diff
@@ -46,17 +46,35 @@ class Chain {
     await this.blockInfo.update();
   }
 
+  nameAsPascalCase() {
```
antonina-cherednichenko, line 49: maybe just uppercase for all chain names? definitely it looks exotic right now

toninorair (approved):
Great work, I was a little confused why most of the names involve Matic, but there are still some with Polygon, other than that looks good.

ethereum-client/src/events.rs
```diff
@@ -312,6 +312,8 @@ pub fn decode_event(topics: Vec<String>, data: String) -> Result<EthereumEvent,
         .iter()
         .map(|topic| decode_topic(topic).ok_or(EventError::InvalidTopic))
         .collect::<Result<Vec<ethabi::Hash>, _>>()?;
+
+    // xxx todo:wn note - this is a strange match - need to check the syntax on this and see what it really means.....
```
jflatow: You mean using the guards?

pallets/cash/src/chains.rs
```diff
@@ -25,13 +25,15 @@ pub enum ChainId {
...
     Dot,
+    Matic, // xxx todo: due to rebrand ticker and name don't match for MATIC/Polygon.. so review this enum name is this what we want to call this?
 }
 
 impl ChainId {
     pub fn to_account(self, addr: &str) -> Result<ChainAccount, Reason> {
         match self {
             ChainId::Gate => Ok(ChainAccount::Gate(Gateway::str_to_address(addr)?)),
             ChainId::Eth => Ok(ChainAccount::Eth(Ethereum::str_to_address(addr)?)),
+            ChainId::Matic => Ok(ChainAccount::Matic(Polygon::str_to_address(addr)?)),
```
coburncoburn, line 36: these should be append only, if Dot has ever been written to storage, it will be wrong next time it is read after this patch. We'd probably get away with it this time, but it's a huge pain if this get's messed up

pallets/cash/src/chains.rs
```diff
@@ -25,13 +25,15 @@ pub enum ChainId {
     Gate,
     Eth,
     Dot,
+    Matic, // xxx todo: due to rebrand ticker and name don't match for MATIC/Polygon.. so review this enum name is this what we want to call this?
```
jflatow: Yeah I think this makes sense, not sure what else we would call the network and this still seems to be what they are frequently calling it

pallets/cash/src/chains.rs
```diff
@@ -348,72 +394,138 @@ impl FromStr for ChainId {
 #[derive(Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug, Types)]
 pub enum ChainBlock {
     Eth(<Ethereum as Chain>::Block),
+    Matic(<Polygon as Chain>::Block),
```
coburncoburn, line 398: kind of funny to have call it Matic and Polygon on the same line 🤔

pallets/cash/src/chains.rs
```diff
@@ -348,72 +394,138 @@ impl FromStr for ChainId {
...
+        let return_value: Box<dyn Iterator<Item = ChainBlockEvent>> = match self {
+            ChainBlock::Eth(block) => {
+                Box::new(Self::map_events_eth_like(block, ChainBlockEvent::Eth))
+            }
+            ChainBlock::Matic(block) => {
+                Box::new(Self::map_events_eth_like(block, ChainBlockEvent::Matic))
+            }
+        };
+
+        return_value
```
jflatow: Yeah its an extra allocate but as a `Vec<ChainBlockEvent>` should be less line noise too

pallets/cash/src/chains.rs
```diff
@@ -348,72 +394,138 @@ impl FromStr for ChainId {
...
+
+    fn blocks_numbers_eth_like(blocks: &Vec<EthereumBlock>) -> impl Iterator<Item = u64> + '_ {
+        blocks.iter().map(|b| b.number)
     }
 
     pub fn block_numbers(&self) -> impl Iterator<Item = u64> + '_ {
-        match self {
-            ChainBlocks::Eth(blocks) => blocks.iter().map(|b| b.number),
-        }
+        let return_value: Box<dyn Iterator<Item = u64>> = match self {
```
jflatow: Hm actually surprised it even complains about this one since they are always `u64` items

pallets/cash/src/core.rs
```diff
@@ -319,6 +330,57 @@ pub fn apply_chain_event_internal<T: Config>(event: &ChainBlockEvent) -> Result<
                 result.to_vec(),
             ),
         },
+        ChainBlockEvent::Matic(_block_num, eth_event) => match eth_event {
+            ethereum_client::EthereumEvent::Lock {
```
coburncoburn, line 334: maybe these should be refactored to `EthLikeEvent` ?

pallets/cash/src/core.rs
```diff
@@ -354,6 +416,34 @@ pub fn unapply_chain_event_internal<T: Config>(event: &ChainBlockEvent) -> Resul
...
+            _ => Ok(()),
+        },
+        ChainBlockEvent::Matic(_block_num, eth_event) => match eth_event {
+            ethereum_client::EthereumEvent::Lock {
+                asset,
+                sender,
+                chain,
+                recipient,
+                amount,
+            } => internal::lock::undo_lock_internal::<T>(
```
coburncoburn, line 428: tbh this boiler plate is a bit unnerving.  copy pasting these same calls for each chain is kind of scary

pallets/cash/src/chains.rs
waynenilsen: yeah there is a reallly strange error on that one, that is actually the case with a lot of these where the item is actually the same exact type all the way down `EthereumBlock` for example, I couldn't really make sense of the explanation at the time from what i remember. Honestly though this is a very small allocation relative to an entire Vec. Although, they're all quite small allocations overall I guess.. I suppose it is possible that some long lists of blocks could overflow the stack though

pallets/cash/src/chains.rs
waynenilsen: yeah "matic" is the ticker, "matic" is to Polygon as "eth" is to Ethereum

ethereum-client/src/events.rs
waynenilsen: Ok i just did the todo. It turns out that the thing to the left of the `=>` must be a `PATTERN` and here are the docs for this kind of pattern, it is called a "Match Guard".

https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#extra-conditionals-with-match-guards

I _thought_ `t if *t == *LOCK_EVENT_TOPIC` was being evaluated as an expression somehow with that expressions output resulting in the LHS of the match. Importantly `t if *t == *LOCK_EVENT_TOPIC` is NOT in fact an expression and can not be.

coburncoburn:
Looks like a 🧼 implentation of the current architecture, but going through this makes me think we should pre process the blocks and events higher up in the pipeline so more of the runtime code is dealing with `StarportEvent` rather than `EthLikeEvent`, for example. Any thoughts on coming up with an interface for `StarportBlock` etc rather than `EthBlock, FlowBlock`? We should be able to the the branching one time early rather than over and over again.

jflatow:
> Looks like a 🧼 implentation of the current architecture, but going through this makes me think we should pre process the blocks and events higher up in the pipeline so more of the runtime code is dealing with `StarportEvent` rather than `EthLikeEvent`, for example. Any thoughts on coming up with an interface for `StarportBlock` etc rather than `EthBlock, FlowBlock`? We should be able to the the branching one time early rather than over and over again.

I avoided trying to generalize across the chains since most of them we haven't seen yet, and really they are independent. We'll definitely need improvements for 'superblocks' and things at some point, but these kind of *are* the interfaces, just exposed as functions right? Definitely worth discussing

waynenilsen:
Yeah when I was writing the code I was pretty unsatisfied with amount of boilerplate but generally speaking I wanted to implement this feature without major refactoring. 

I think that when we see how Flow fits into the code refactoring will become a bit more obvious and because it is not "eth like" we will see what ends up being actually generic/common and what just seems generic because eth and matic are so darn similar.

pallets/cash/src/core.rs
waynenilsen: Yeah the same thing could probably be said for `ethereum_client` and all of the other things in there EthereumBlock etc.. not sure if that makes the situation clearer or not though

coburncoburn:
> > Looks like a 🧼 implentation of the current architecture, but going through this makes me think we should pre process the blocks and events higher up in the pipeline so more of the runtime code is dealing with `StarportEvent` rather than `EthLikeEvent`, for example. Any thoughts on coming up with an interface for `StarportBlock` etc rather than `EthBlock, FlowBlock`? We should be able to the the branching one time early rather than over and over again.
> 
> I avoided trying to generalize across the chains since most of them we haven't seen yet, and really they are independent. We'll definitely need improvements for 'superblocks' and things at some point, but these kind of _are_ the interfaces, just exposed as functions right? Definitely worth discussing

I think the first time we went through this it was more like "ChainThing.do_something" where ChainThing is an trait that EthThing and FlowThing implement and we pass around ChainThing with [dyn](https://doc.rust-lang.org/reference/types/trait-object.html), which would be very close to the enum branches in reality, possible overhead of Boxes and dyn not being *that* cool. 

I'm thinking this time instead we could make gateway very very agnostic of L1's like if the OCW pushes a single type `StarportEvent` into Gateway rather than one of n enum in `ChainEvent`. 

Agreed fits best into superblock discussion

pallets/cash/src/core.rs
```diff
@@ -240,6 +243,14 @@ pub fn recover_validator<T: Config>(
                 }
             }
         }
+        ChainSignature::Matic(eth_sig) => {
```
toninorair, line 246: will validator only really produce ChainSignature::Eth? do we really need to add more chains here?

toninorair: @waynenilsen this most likely should go, here is the discussion https://discord.com/channels/402910780124561410/796451735803002900/857390332396568596

pallets/runtime-interfaces/src/lib.rs
```diff
@@ -40,12 +43,14 @@ pub fn initialize_validator_config(
...
 const ETH_KEY_ID_ENV_VAR: &str = "ETH_KEY_ID";
 const ETH_RPC_URL_ENV_VAR: &str = "ETH_RPC_URL";
+const MATIC_RPC_URL_ENV_VAR: &str = "MATIC_RPC_URL";
 const MINER_ENV_VAR: &str = "MINER";
 const OPF_URL_ENV_VAR: &str = "OPF_URL";
 
 const ETH_KEY_ID_DEFAULT: &str = gateway_crypto::ETH_KEY_ID_ENV_VAR_DEV_DEFAULT;
 const MINER_DEFAULT: &str = "Eth:0x0000000000000000000000000000000000000000";
 const ETH_RPC_URL_DEFAULT: &str = "https://ropsten-eth.compound.finance";
+const MATIC_RPC_URL_DEFAULT: &str = "https://ropsten-eth.compound.finance"; // xxx todo:wn fix this before merge
```
toninorair: Looks like this needs a fix, especially giving the comment

waynenilsen: Yeah, these defaults aren't really terribly useful anyway. I think I'll just set it to `""` and call it a day.

pallets/cash/src/core.rs
waynenilsen: when i took another look at this, i completely agree with you, i _think_ I had to put this in there due to the mechanics of the rest of the code (for now) but it was really not what i wanted to do. I'm going to check out what happens if I take it back out but this is a great call out 

I think the other side of the link is here 

`chains.rs:353`

```
pub fn has_validator_signature(&self, chain_id: ChainId, validator: &ValidatorKeys) -> bool {
        match chain_id {
            ChainId::Eth => self.has_signer(ChainAccount::Eth(validator.eth_address)),
            ChainId::Matic => self.has_signer(ChainAccount::Matic(validator.eth_address)),
            _ => false,
        }
    }
```

need to give this some more thought though for sure, it is something that I struggled with in this PR

waynenilsen: also this and various other places 

`chains.rs:361`

```
    pub fn add_validator_signature(
        &mut self,
        signature: &ChainSignature,
        validator: &ValidatorKeys,
    ) -> Result<(), Reason> {
        match (self, signature) {
            (ChainSignatureList::Eth(eth_sig_list), ChainSignature::Eth(eth_sig)) => {
                Ok(eth_sig_list.push((validator.eth_address, eth_sig.clone())))
            }
            (ChainSignatureList::Matic(eth_sig_list), ChainSignature::Matic(eth_sig)) => {
                Ok(eth_sig_list.push((validator.eth_address, eth_sig.clone())))
            }
            _ => Err(Reason::SignatureMismatch),
        }
    }
```

definitely need to think about how we want to deal with this.

Basically, right now, i'm just re-using the same keys for eth and matic since their curve and signing situation is compatible `eth_address` may be something of a misnomer, maybe it should be something more like `eth_like_address` or `evm_compatible_address` etc

integration/util/scenario/starport.js
```diff
@@ -21,8 +21,7 @@ class Starport {
   }
 
   chainAddressStr() {
-    // xxx todo:wn parameterize by chain
-    return `ETH:${this.ethAddress()}`;
+    return `${this.chain.name.toUpperCase()}:${this.ethAddress()}`;
```
jflatow, line 24: Still not generalized though, right?

integration/util/scenario/starport.js
```diff
@@ -202,21 +201,21 @@ async function buildStarport(starportInfo, validatorsInfoHash, ctx, chain, cashT
...
     let proxy = await chain.__deploy('TransparentUpgradeableProxy', [
       starportImpl._address,
       proxyAdmin._address,
       "0x"
-    ], { from: chain.root() });
+    ], {from: chain.root()});
     let starport = await chain.__getContractAt('Starport', proxy._address);
     if (validators.length > 0) {
-      await starport.methods.changeAuthorities(validators).send({ from: chain.root(), gas: 4_000_00 });
+      await starport.methods.changeAuthorities(validators).send({from: chain.root(), gas: 4_000_00});
```
jflatow: Is this gas amount missing a `0`?

integration/util/scenario/token.js
```diff
@@ -160,6 +167,10 @@ class Token {
   async totalChainBorrows() {
     return this.toTokenAmount(await this.ctx.getApi().query.cash.totalBorrowAssets(this.toChainAsset()));
   }
+
+  async getChainBalance(actor) {
+
```
hayesgm: Unused?

integration/util/scenario/starport.js
waynenilsen: Yeah starport.js is quite "eth like" specific right now, that refactor will most likely come with flow

integration/util/scenario/starport.js
waynenilsen: hmm it does look that way but i actually only accidentally changed whitespace in here, wondering if I should change the underscores or add the zero, for now i'll change the underscore so it doesn't impact functionality (which is working)

ethereum/contracts/Starport.sol
```diff
@@ -35,9 +35,23 @@ contract Starport {
...
+    constructor(ICash cash_, address admin_, bytes32 chainId_, bytes4 chainIdHeader_) {
         cash = cash_;
         admin = admin_;
+        chainId = chainId_;
+        chainIdHeader = chainIdHeader_;
+    }
+
+    function bytes32ToString(bytes32 _bytes32) public pure returns (string memory) {
+        uint8 i = 0;
+        while(i < 32 && _bytes32[i] != 0) {
```
hayesgm: `while<space>(`

integration/util/scenario/actor.js
```diff
@@ -32,11 +32,13 @@ class Actor {
   }
 
   toChainAccount() {
-    return { Eth: this.ethAddress() };
+    const returnValue = {};
+    returnValue[this.chain().nameAsPascalCase()] = this.ethAddress();
```
hayesgm, line 36: this might be `localChainAddress()` instead of `ethAddress()`, possibly?

integration/util/scenario/actor.js
```diff
@@ -138,19 +142,28 @@ class Actor {
     return Number(balance) / 1e6;
   }
 
-  async liquidityForToken(token) {
+  async balanceForToken(token) {
     let assetBalance = await this.ctx.getApi().query.cash.assetBalances(token.toChainAsset(), this.toChainAccount());
+    if(assetBalance == 0) {
```
hayesgm: `if<space>` and `===` to prevent type coercion.

integration/__tests__/lock_scen.js
```diff
@@ -6,7 +6,8 @@ let lock_scen_info = {
   tokens: [
     { token: 'usdc', balances: { ashley: 1000 } }
   ],
-  validators: ['alice', 'bob']
+  validators: ['alice', 'bob'],
+  native: true
```
hayesgm, line 10: We don't need native here, but I'm not strictly against it, either. Probably should just use the default though, nah?

integration/__tests__/matic_scen.js
```diff
@@ -1,31 +1,120 @@
...
+            await darlene.lock(10000, maticZrx);
+            balance = await darlene.balanceForToken(maticZrx);
+            let liquidity1 = await darlene.liquidity();
+            expect(balance).toEqual(10000);
+            expect(liquidity1).toBeGreaterThan(0);
+            // now we know everything is in order, make the transfer, let's say half our liquidity worth
+            await darlene.transfer(1000, usdc, chuck);
+            expect(await darlene.chainBalance(usdc)).toEqual(-1000);
+            expect(await chuck.chainBalance(usdc)).toEqual(1000);
+            // xxx todo:wn this test is broken but so is its friend, interest is accruing even though time is not "on"
```
hayesgm: We should fix this, right?

hayesgm: Also, we might want to actually account for interest?

integration/__tests__/matic_scen.js
```diff
@@ -1,31 +1,120 @@
...
             name: 'matic',
-            provider: 'ganache', // [env=PROVIDER]
+            provider: 'ganache',
             ganache: {
                 opts: {},
                 web3_port: null
             },
         },
     },
+    native: true, // for log messages - removes wasm:stripped - no need for wasm no forkless upgrade testing here
```
hayesgm, line 25: This is more for freeze time here. You should have log messages so long as you run `yarn build` which handles the `wasm stripped` concerns already.

integration/util/scenario/actor.js
```diff
@@ -138,19 +142,28 @@ class Actor {
...
+
+    return token.toTokenAmount(assetBalance.toBigInt());
+  }
+
+  async liquidityForToken(token) {
+    let assetBalance = await this.balanceForToken(token);
     let price = await token.getPrice();
     let liquidityFactor = await token.getLiquidityFactor();
 
     if (assetBalance == 0) {
```
hayesgm: This one should also be `===`

integration/util/scenario/chains.js
```diff
@@ -47,17 +47,44 @@ class Chain {
...
+
   genesisBlock() {
-    return {
-      Eth: {
-        hash: this.blockInfo.hash,
-        parentHash: this.blockInfo.parentHash,
-        number: this.blockInfo.number,
-        events: []
-      }
+    const block = {};
```
hayesgm:
```js
return {
  [this.nameAsPascalCase()]: {
    hash: this.blockInfo.hash,
    // ...
  }
}```

File: integration/util/scenario/chains.js [hunk 25]
```diff
@@ -47,17 +47,44 @@ class Chain {
...
+      hash: this.blockInfo.hash,
+      parentHash: this.blockInfo.parentHash,
+      number: this.blockInfo.number,
+      events: []
     };
+    return block;
   }
 
+  genesisBlockForChainSpec() {
+    const newBlock = {};
```
hayesgm: Can do a similar `{ [xxx]: yyy }` pattern here for nicer syntax.

integration/util/scenario/chains.js
```diff
@@ -233,6 +260,10 @@ class Chains {
       deployment.chain.cashToken = deployment.cashToken;
     });
   }
+
+  genesisBlocksForChainSpec() {
+    return this.all().map(chain => chain.genesisBlockForChainSpec());
```
hayesgm, line 264: This returns an array of `[ { Eth: {}, Matic: {} ]`, right?

integration/util/scenario/chains.js
```diff
@@ -272,7 +303,7 @@ async function buildChain(chainInfo, ctx) {
   let version = chainInfo.version ? ctx.versions.mustFind(chainInfo.version) : ctx.versions.current;
 
   let blockInfo = new BlockInfo(web3, chainInfo.block_number, ctx);
-  await blockInfo.update();
+  await blockInfo.update('latest');
```
hayesgm: Little confused on latest here, doesn't this have a `block_number` that may or may not be latest?

integration/util/scenario/starport.js
```diff
@@ -201,21 +200,21 @@ async function buildStarport(starportInfo, validatorsInfoHash, ctx, chain, cashT
...
     if (validators.length > 0) {
-      await starport.methods.changeAuthorities(validators).send({ from: chain.root(), gas: 4_000_00 });
+      await starport.methods.changeAuthorities(validators).send({from: chain.root(), gas: 400_000});
     }
 
     let starportTopics = Object.fromEntries(starport
-      ._jsonInterface
-      .filter(e => e.type === 'event')
-      .map(e => [e.name, e.signature]));
+        ._jsonInterface
```
hayesgm: I think we're using 2-spaces here?

pallets/cash/src/chains.rs
```diff
@@ -1004,18 +1265,28 @@ impl Chain for Polkadot {
...
 
 pub fn get_chain_account(chain: String, recipient: [u8; 32]) -> Result<ChainAccount, Reason> {
-    match &chain.to_ascii_uppercase()[..] {
-        "ETH" => {
-            let mut eth_recipient: [u8; 20] = [0; 20];
-            eth_recipient[..].clone_from_slice(&recipient[0..20]);
+    let chain_account_fn = match &chain.to_ascii_uppercase()[..] {
+        "ETH" => ChainAccount::Eth,
+        "MATIC" => ChainAccount::Matic,
+        "MAT" => ChainAccount::Matic,
```
hayesgm, line 1286: Curious about the either here.

integration/__tests__/lock_scen.js
waynenilsen: this was before we were doing wasm logging was impossible to tell what was going on, also is native faster?

pallets/cash/src/chains.rs
waynenilsen: this is the whole `bytes4` thing :( i use `MAT:` for the magic header and MATIC for addresses etc. Do you want me to rename the other ones to `MAT` just seemed odd since `eth` is the ticker for Ethereum and `MATIC` is the ticker for Polygon

We can really name these whatever we want but i think it might be nice to use the tickers since they're already defacto market imposed short names on each chain we might bridge

integration/util/scenario/actor.js
waynenilsen: Yeah, I think for now having ethAddress is ok but we will ultimately change a lot of this code when flow comes into play. To be honest, I'm a little bit intimidated refactoring something as pervasive as that could be in js without the help of a compiler

integration/__tests__/matic_scen.js
waynenilsen: Yeah I can work on this test and its friend

integration/util/scenario/chains.js
waynenilsen: ah thank you!

integration/util/scenario/chains.js
waynenilsen: Yes

integration/util/scenario/chains.js
waynenilsen: good find 🦅 👁️ this is a vestigial parameter

integration/__tests__/matic_scen.js
waynenilsen: TL;DR - it turns out that the problem is that I was building without the `integration` feature. 

I was doing this because I generally preferred to use debug profile. I mostly have this preference because as I work on rust code I often `cargo build` when i'm ready. If I then `cargo build --release` as in `yarn build` it ends up having to build everything again from scratch. 

Compile times are pretty painful which lead me down this road generally. Cargo check isn't too much better because ultimately you're going to want a build and when you're ready for a build it needs to do the process from scratch anyway. At least rust has incremental builds so that when you don't touch a lot of the code it doesn't have to rebuild much.

integration/__tests__/lock_scen.js
hayesgm: TBH, it doesn't really matter here. Native is when you're asking for a different genesis native, e.g. running m14's native build. It's irrelevant here, but also, again, if we wanted to change it, we should change the default, not one one-off scen.

integration/util/scenario/actor.js
hayesgm: `Cmd+Shift+F` and JavaScript are sadly good friends.

hayesgm: FWIW, we could use TypeScript, but it's a huge burden for tests since it takes a lot more time to write and tests are pretty one-offy. We could technically upgrade if we think it would help us.

PR #333, opened 2021-05-20
Update the integration testing readme

waynenilsen:
[rendered](https://github.com/compound-finance/gateway/blob/wayne/update-integration-test-readme/integration/README.md)

### DIFF
--- a/integration/README.md
+++ b/integration/README.md
@@ -3,42 +3,39 @@
 
 Integration tests that run the real deal and make sure every component works together. The goal of these tests are to show that end-to-end flows are correctly working. Specifically, we will run ganache, deploy Cash Token, Starport, etc, run a simplified Gateway test-net and then interact with ganache and Gateway to say mint some Cash or extract collateral from Gateway. These are fully automated and run in CI.
 
+## Building
+
+You can build the Starport code and Gateway node using:
+
+```sh
+gateway/integration> yarn && yarn build
+```
+
+This will build the chain using `release` mode, and special flags used for integration testing.
+
 ## Running
 
-First, you'll need to compile Gateway (in release mode) and compile the Ethereum contracts.
+Running the entire test suite doesn't really work right now so run one test at a time by using `only:true` on the test you want to run as follows:
 
-```sh
-gateway> cargo +nightly build
+```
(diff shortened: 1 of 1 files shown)


jflatow:
Cool, thanks for trying to keep this up to date

integration/README.md
```diff
@@ -5,23 +5,23 @@ Integration tests that run the real deal and make sure every component works tog
 
 ## Running
 
-First, you'll need to compile Gateway (in release mode) and compile the Ethereum contracts.
+You can run the tests with Gateway compiled in release or debug profile. The default profile is **debug**.
```
hayesgm: This is not true, at least not anymore. Really now we should use `yarn build` and `yarn test` which will do all of the right things all the time.

integration/README.md
```diff
@@ -5,23 +5,23 @@ Integration tests that run the real deal and make sure every component works tog
...
 ```sh
 gateway> cargo +nightly build
 ```
 
-Note: if you require deeper debugging, you may want to enable the `runtime-debug` feature, via:
+Note: if you require deeper debugging, you may want to enable the `integration` feature, via:
 
 ```sh
-gateway> cargo +nightly build --features "runtime-debug"
+gateway> WASM_BUILD_RUSTFLAGS='--cfg feature=\"integration\"' cargo build --release --features integration
```
hayesgm: I feel like these will be fragile versus using `yarn build` from `integration/`

integration/README.md
```diff
@@ -33,12 +33,34 @@ Next, install integration test dependencies in this directory:
...
+```
+
+If you built with the `--release` flag, then run the test using: 
 
 ```sh
-yarn test
+yarn test __tests__/<my-test>.js
 ```
 
+If you did NOT build with the `--release` flag, then run the test using:
```
hayesgm: You can actually just do `PROFILE=debug yarn test ...` I believe.

hayesgm (changes requested):
Awesome thanks; one thing in regards to the new default being `release` instead of debug. Also, possibly just focus on using the `yarn XYZ` variants since they can be kept up-to-date a little easier.

PR #334, opened 2021-05-20
wip MATIC/Polygon integration testing

waynenilsen:
Note: change the base branch back to develop when we are ready. This is just to show the diff better.

This testing approach uses the same ganache instance and the same contract code but we deploy two of everything. This allows us to simulate being on a different EVM compatible blockchain. The same actors / owners are used but this should not really cause any problem.  The one open question that I have is around how we want to handle token balances. Right now, the token balance is set on both matching tokens so given this code:

```
let lock_scen_info = {
    tokens: [
        { token: 'usdc', balances: { ashley: 1000 } }
    ],
    validators: ['alice', 'bob']
};
```

The ashley address will have 1000 usdc on both chains. That is not ideal but it is now a question of how we want to paramaterize this configuration to say where the balance reside. We could do something like an optional dict with default as in what is above is valid and will default to the ethereum chain and if you want something on a different chain you have to specify the chain name as below

```
let lock_scen_info = {
    tokens: [
        { token: 'usdc', balances: { ashley: {'matic': 1000} } }
    ],
    validators: ['alice', 'bob']
};
```

I admit that this would be preferable 

```
let lock_scen_info = {
    tokens: [
        { chain: 'matic', token: 'usdc', balances: { ashley: {'matic': 1000} } }
    ],
    validators: ['alice', 'bob']
};
```

but that would somehow seem to indicate that the entire token should _only_ exist on that chain which I dont _think_ we want. Perhaps it is something like 

```
let lock_scen_info = {
    tokens: [
        { chains: ['eth','matic'], token: 'usdc', balances: { ashley: {'matic': 1000} } }
    ],
    validators: ['alice', 'bob']
};
```

still a work in progress

### DIFF
--- a/integration/__tests__/matic_scen.js
+++ b/integration/__tests__/matic_scen.js
@@ -1,0 +1,17 @@
+const { buildScenarios } = require('../util/scenario');
+
+let lock_scen_info = {
+    tokens: [
+        { token: 'usdc', balances: { ashley: 1000 } }
+    ],
+    validators: ['alice', 'bob']
+};
+
+buildScenarios('Matic', lock_scen_info, [
+    {
+        name: 'Matic',
+        scenario: async ({ ashley, usdc, chain }) => {
+            expect(true);
+        }
+    },
+]);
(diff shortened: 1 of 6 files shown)


waynenilsen:
closing for #341

PR #336, opened 2021-05-20
m9, The Lesser

hayesgm:
This patch adds storage that will be used in m10. We allow setters to set the values so that we can ensure test-net has the proper values by the time it upgrades to m10. We do nothing more.


jflatow (approved):
LGTM

hayesgm:
force pushed to develop

PR #337, opened 2021-05-21
Fix chain cash regression (do not modify on transfer)

jflatow:


pallets/cash/src/pipeline.rs
```diff
@@ -369,28 +369,30 @@ fn prepare_augment_cash<T: Config>(
...
     let total_cash_post = st
         .get_total_cash_principal::<T>()
         .sub(recipient_repay_principal)
         .map_err(|_| Reason::InsufficientChainCash)?;
 
     st.set_cash_principal::<T>(recipient, recipient_cash_post);
     st.set_total_cash_principal::<T>(total_cash_post);
-    st.set_chain_cash_principal::<T>(chain_id, chain_cash_principal_post);
+
+    if from_external {
```
hayesgm, line 388: Are we allowed to bail after we set things here? Is that behavior used elsewhere?

hayesgm (approved):
Looks good; I think it's relatively clean and fixes a key bug. Quick question in re: semi-stateful effects.

hayesgm (approved):
lgtm

pallets/cash/src/pipeline.rs
jflatow: We don't write until commit right?

jflatow: Or not sure what you mean

PR #338, opened 2021-05-21
Minor m10 Tweaks

hayesgm:
This just patches a few small topics for m10. Nothing serious in the scheme of things.


integration/util/jest.js
```diff
@@ -97,7 +97,10 @@ expect.extend({
   toEthRevert(actual, reason = null) {
     if (reason === null) {
       return {
-        pass: !!actual['message'] && actual.message.startsWith(`Transaction has been reverted by the EVM:`),
+        pass: !!actual['message'] && (
+          actual.message.startsWith(`Transaction has been reverted by the EVM:`)
```
hayesgm, line 101: These messages change on me daily-- I'm not sure why but whatever, they both mean the same thing

PR #341, opened 2021-05-25
refactor testing code to support multiple L1s

waynenilsen:
* includes new `deployments` that packages starport and cashToken together due to their circular dependence
* uses existing patterns to strap ethStarport, ethCashToken on as well as supporting the old `starport` and `cashToken` properties
* should be able to include a new `matic` key in the `chain_opts` to support matic

### DIFF
--- a/integration/util/scenario/actor.js
+++ b/integration/util/scenario/actor.js
@@ -6,12 +6,17 @@ const { lookupBy } = require('../util');
 const { CashToken } = require('./cash_token');
 
 class Actor {
-  constructor(name, ethAddress, chainKey, ctx) {
+  constructor(name, ethAddress, chainKey, chainName, ctx) {
     this.name = name;
     this.__ethAddress = ethAddress;
     this.chainKey = chainKey;
+    this.chainName = chainName;
     this.ctx = ctx;
     this.nextId = 0;
+  }
+
+  chain() {
+    return this.ctx.chains.find(this.chainName());
   }
 
   show() {
@@ -162,7 +167,7 @@ class Actor {
     };
 
     return await this.declare("lock", [amount, asset], async () => {
-      let tx = await this.ctx.starport.lock(this, amount, asset);
+      let tx = await asset.chain().starport.lock(this, amount, asset);
       let event;
       if (opts.awaitEvent) {
         event = await this.ctx.chain.waitForEthProcessEvent('cash', asset.lockEventName());
@@ -180,7 +185,7 @@ class Actor {
       ...opts
     };
     return await this.declare("lock", [amount, asset, "to", recipient], async () => {
-      let tx = await this.ctx.starport.lockTo(this, amount, asset, 'ETH', recipient);
+      let tx = await asset.chain().starport.lockTo(this, amount, asset, 'ETH', recipient);
       let event;
       if (opts.awaitEvent) {
         event = await this.ctx.chain.waitForEthProcessEvent('cash', asset.lockEventName());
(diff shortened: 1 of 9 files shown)


integration/util/scenario/actor.js
```diff
@@ -275,15 +276,24 @@ function actorInfoMap(keyring) {
...
+      chain: 'matic',
+      key_uri: '//Darlene'
     }
   };
 }
 
 async function buildActor(actorName, actorInfo, keyring, index, ctx) {
-  let ethAddress = ctx.eth.accounts[index + 1];
+  let chain = ctx.eth;
+  if(actorInfo.chain) {
```
hayesgm: `if (`

integration/util/scenario/actor.js
```diff
@@ -275,15 +276,24 @@ function actorInfoMap(keyring) {
...
+    darlene: {
+      chain: 'matic',
+      key_uri: '//Darlene'
     }
   };
 }
 
 async function buildActor(actorName, actorInfo, keyring, index, ctx) {
-  let ethAddress = ctx.eth.accounts[index + 1];
+  let chain = ctx.eth;
```
hayesgm: I wonder if chain is better as `chainName: 'eth'` and then a function `actor.chain() { return this.ctx.chains.find(this.chainName) }`. This isn't that important, but it could make it easy to not end up copying an object that is otherwise stateful and thus gets confusing when you modify it and it doesn't appear modified by certain references (since they are copied).

integration/util/scenario/chains.js
```diff
@@ -230,15 +253,37 @@ async function buildEth(ethInfo, ctx) {
...
 
-  return new Eth(ethInfo, web3, web3Url, accounts, blockInfo, ganacheServer, version, ctx);
+  return new Chain(chainInfo, web3, web3Url, accounts, blockInfo, ganacheServer, version, ctx);
+}
+
+async function buildChains(chainsInfo, ctx) {
+  const promises = {};
+  const chains = {};
+
+  // start chains in parallel
```
hayesgm: yeah, there's a good function `await Promise.all(promises)` for handling this type of issue.

integration/util/scenario/ctx.js
```diff
@@ -298,15 +299,21 @@ async function buildCtx(scenInfo={}) {
   ctx.log(`Building ctx with scenInfo=${JSON.stringify(scenInfo, null, 2)}`);
   ctx.log(`test=${JSON.stringify(scenInfo.chain_spec, null, 2)}`);
   ctx.versions = await buildVersions(scenInfo.versions, ctx);
-  ctx.eth = await buildEth(scenInfo.eth_opts, ctx);
+  ctx.chains = await buildChains(scenInfo.chain_opts, ctx);
+  aliasBy(ctx, ctx.chains.all(), 'name');
 
   // Note: `3` below is the number of transactions we expect to occur between now and when
```
hayesgm: This comment should be moved to `deployments.js:36`

integration/util/scenario/starport.js
```diff
@@ -165,33 +169,35 @@ class Starport {
...
-async function buildStarport(starportInfo, validatorsInfoHash, ctx) {
-  ctx.log("Deploying Starport...");
-  if (!ctx.cashToken) {
-    throw new Error(`Cannot deploy Starport without first deploying Cash Token`);
-  }
+async function buildStarport(starportInfo, validatorsInfoHash, ctx, chain, cashToken) {
+  ctx.log(`Deploying Starport to ${chain.name}...`);
+
+  // xxx todo:wn add back sanity check (?)
+  // if (!ctx.cashToken) {
```
hayesgm: I assume this check was there since we _needed it_. If it's not required, we can probably remove it.

integration/util/scenario/token.js
```diff
@@ -205,6 +206,7 @@ function tokenInfoMap(ctx) {
 
   return {
     zrx: {
+      chain: 'matic',
```
hayesgm, line 223: We do use `zrx` in tests, so we probably should make a new token / set of tokens for matic.

hayesgm (approved):
Looks _perfect_. Few minor comments.

integration/util/scenario/chains.js
waynenilsen: changed it to this, not sure how much clearer it is but there are fewer lines 

```
  const chainsList = await Promise.all(
      Object.values(chainsInfo)
          .map(chainInfo => buildChain(chainInfo, ctx))
  )
  // maybe a clearer/easier way to do that
  const chains = chainsList.reduce((acc, chain) => {acc[chain.name] = chain; return acc;}, {})
```

integration/util/scenario/token.js
waynenilsen: found a problem with `setSupplyCap` when I addressed this so, yay :)

PR #342, opened 2021-05-26
Flow Support

toninorair:
Adding support for the Flow blockchain. The following PR includes:
1) Starport implementation on the Flow blockchain, transactions and scripts for interacting with it, unit tests, and fcl files for UI interaction with Starport.  

> Starport deployment https://flow-view-source.com/testnet/account/0xc8873a26b148ed14/contract/Starport

> Starport documentation https://docs.google.com/document/d/16Y3e2klvUOKqWm8jMOUDJZihx6SWnO0Gg2cRr2g2i38/edit

> Example of Flow tokens `Lock` transaction 
https://flow-view-source.com/testnet/tx/3410f92c5818cdfab9eaa3fc9e236359f7d269b34c72d106d0ec853ba79e83a3/

2) Flow client for the Gateway (similar to Ethereum client)
3) Flow events fetcher - Go server that utilizes GO SDK to fetch events and blocks from Go testnet/mainnet.

The following PR DOES NOT INCLUDE:
Integration of flow client to Gateway Cash Pallet and scenario tests (coming in the next PR).


flow/flow-fetcher/events.go
```diff
@@ -0,0 +1,70 @@
...
+	Data             string `json:"data"`
+}
+
+// type LockData struct {
+// 	Asset     string  `json:"asset"`
+// 	Recipient string  `json:"recipient"`
+// 	Amount    float64 `json:"amount"`
+// }
+
+func getLockEvents(flowClient *client.Client, eventsInfo FlowEventsInfo) ([]FlowEvent, error) {
```
waynenilsen, line 28: Is this more generic than merely `getLockEvents`?

flow/flow-fetcher/events.go
```diff
@@ -0,0 +1,70 @@
...
+
+			// Build Lock event result data
+			var eventRes = FlowEvent{
+				BlockId:          blockEvent.BlockID.String(),
+				BlockHeight:      blockEvent.Height,
+				TransactionId:    lockEvent.TransactionID.String(),
+				TransactionIndex: lockEvent.TransactionIndex,
+				EventIndex:       lockEvent.EventIndex,
+				Topic:            lockEvent.Type,
+				Data:             fmt.Sprintf("{\"asset\":\"%s\",\"recipient\":\"%s\",\"amount\":%d}", event.Asset(), event.Recipient(), event.Amount()),
```
waynenilsen, line 57: Possible double encode here? Does this end up being the whole escaped json string inside a json payload type of thing?

flow/flow-fetcher/starport/flow_lock_event.go
```diff
@@ -0,0 +1,34 @@
+package starport
```
waynenilsen, line 1: Generally throughout this file should we be using checked access with err return values? What happens when you do `evt.Fields[0]` but `len(evt.Fields) == 0` ? Is that a go panic at that point? Same goes for unchecked casts in this file

flow/flow-fetcher/blocks.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func getBlock(flowClient *client.Client, blockInfo FlowBlockInfo) (Block, error) {
+	// If height and id of block are no set, return the latest block
+	if blockInfo.Id == "" && blockInfo.Height == 0 {
+		return getLatestBlock(flowClient)
+	}
+
+	block, err := func() (*flow.Block, error) {
+		if blockInfo.Height == 0 {
```
waynenilsen, line 46: this is an example of where i think we're perhaps being a little too "smart" in this wrapper? again - total nit

waynenilsen (commented):
Nice work! 🎉  I did not yet look at the cadence part (nor do i really have the expertise to, but who does ;))

1

The whole thing about where to import `from` is pretty sad eg `0x0ae53cb6e3f42a79, 0xee82856bf20e2aa6, ...` We have this in ethereum too to some extent but at least it is parameterized by something like `address xyzContract`. I think this means we really need to think about the deployment process here. Might be worth using a template engine to supply the addresses with some kind of build phase. I'm sure they know this is a problem for anyone building on this system but they just don't have any tooling around it yet :( . 

2 

This is really a pretty small nit. Broadly speaking, I am of the opinion that `flow-fetcher` should be as dumb a wrapper as possible so that as we expand functionality from within the OCW we can access whatever we need from this proxy. I think we are pretty close right now but I think this does a little too much re-shaping that is why this is merely a nit. I think ideally we would just re-expose methods from `flowClient *client.Client` as "RPC" endpoints. 

To me, the endpoints should be something like `POST /GetBlockByID` and `POST /GetBlockByHeight` with json post bodies that match to the args by name and return values that match exactly to their gRPC counterparts. Go being as it is, mapping the return values exactly may require a decent amount of boilerplate but maybe not. I think the objects generated by protobuf serialize to JSON pretty well if I remember correctly. We should be able to serialize those return values directly to JSON.

One downside that I see here is that we somewhat lose control over the return value shapes so that when we update the version of flow in `go.mod` our return values may change shape from under our feet and break our api. Pros and cons I suppose. Maybe that is where we include `/v1` so we can have backward compatible versions?

What do you think?

3

It would be cool to see some unit tests of the `flow-fetcher`. It could be done with mocks or with a real flow client. Go unit testing is pretty friendly if not boilerplate heavy, the easiest way to do it if you want to mock things out is to mostly depend on interfaces then mock out the interfaces. Reminds me a lot of Rust unit testing in that way. The dependency injection story in Go is still woefully lacking and mostly done by hand as far as I can tell.

PR #343, opened 2021-05-27
Attempt to patch issue blocking ingestion

jflatow:
Only filter the blocks that were supported (not dissented), and only allow support or dissent for a given signer.
The vote is being cast only on the support side, dissent is only derived if the chain receives a disagreement with the first one it received.


hayesgm:
This can be closed in lieu of #353

PR #347, opened 2021-06-01
Test m11 Patch

hayesgm:


integration/__tests__/lock_scen.js
```diff
@@ -22,7 +22,18 @@ let now = Date.now();
 
 buildScenarios('Lock Scenarios', lock_scen_info, [
   {
+    only: true,
```
jflatow: Temporary?

integration/util/scenario/proxy.js
```diff
@@ -0,0 +1,116 @@
...
+        let clientBodyP = readRequest(clientReq);
+
+        let options = {
+          method: clientReq.method,
+          headers: clientReq.headers,
+          ...getOptions(that.fallbackUrl)
+        };
+
+        let proxy = http.request(options, async function (res) {
+          await that.hold; // Hold all requests to let us have time to set-up replacements
```
jflatow, line 43: This is pretty gnarly, what's going on with these holds actually?

integration/__tests__/upgrade_to_m10__hardfork_scen.js
```diff
@@ -0,0 +1,185 @@
...
+      let proxyJson = (json) => {
+        return (req, res, body) => {
+          res.setHeader("Content-Type", "application/json");
+          res.writeHead(200);
+          res.end(JSON.stringify(json, null, 4));
+        };
+      };
+
+      [charlie.ethProxy, dave.ethProxy].forEach((proxy) => {
+        proxy.replace(/eth_getBlockByNumber.*\"0x1\"/, proxyJson(badBlock1));
```
jflatow, line 133: This seems like it'll be a bit fragile using a regex on the raw body instead of matching fields, if the order in the JSON changes it'll just silently stop replacing?

hayesgm:
Closing in lieu of #353

PR #348, opened 2021-06-02
Substrate upgrade + Gateway addresses without removing unnecessary variants

jflatow:


jflatow:
We're going to upgrade substrate

PR #350, opened 2021-06-04
Cargo audit fixes

jflatow:
Also upgrade substrate again in order to bump deps. This is using another substrate branch so as not to break the one others are using.

Couldn't update cranelift-codegen without starting to break stuff in substrate.


hayesgm:
I believe this is clear to close under #353

jflatow:
Subsumed by #353

PR #352, opened 2021-06-08
Start to build code substitutes

jflatow:


runtime/src/lib.rs
```diff
@@ -120,8 +120,8 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("gateway"),
     impl_name: create_runtime_str!("gateway"),
     authoring_version: 1,
-    spec_version: 10,
-    impl_version: 2,
+    spec_version: 11,
+    impl_version: 1,
```
jflatow, line 124: Do we still want this?

hayesgm:
This is incorporated already, as far as I can tell.

PR #353, opened 2021-06-08
Upgrade everything we can for a chain reset

jflatow:
This branch is meant to be an integration point for all changes going into the next release, which at this point is looking to be a brand new testnet chain. We lost our ability to upgrade via governance with the re-org breakage, and hard forking has proved difficult, so we need to spend more time on it (but don't want to block people from using the testnet in the meantime).

At the moment, this will still need to be rebased onto develop, which has some other changes to scenarios which may conflict. There are also a few other branches that should get rebased / merged into this one, and then probably a few fixes to make sure CI still works after everything else is merged.


runtime/src/lib.rs
```diff
@@ -119,7 +120,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("gateway"),
     impl_name: create_runtime_str!("gateway"),
     authoring_version: 1,
-    spec_version: 10,
+    spec_version: 11,
```
jflatow, line 126: We still want this?

integration/__tests__/upgrade_to_m10__hardfork_scen.js
```diff
@@ -0,0 +1,185 @@
+const { buildScenarios } = require('../util/scenario');
```
jflatow, line 1: Should we just delete this whole scenario or what?

integration/__tests__/lock_scen.js
```diff
@@ -23,6 +23,16 @@ let now = Date.now();
 buildScenarios('Lock Scenarios', lock_scen_info, [
   {
     name: 'Lock Collateral',
+    info: {
```
hayesgm: I don't think we need this; just wanted to test it here for re-org testing.

integration/__tests__/upgrade_to_m10__hardfork_scen.js
hayesgm: I'd just mark it as skip for now and then delete it separately. Just want to make sure we have something to revive in the future.

integration/babel.config.json
```diff
@@ -0,0 +1,13 @@
+{
```
hayesgm, line 1: I'd like to remove this if I can, but we're stuck with it for now.

integration/util/scenario/ctx.js
```diff
@@ -68,10 +68,13 @@ class Ctx {
...
 
+  __deepColor() {
+    return !!process.env['DEEP_COLOR'];
+  }
+
   __abort(msg) {
-    console.error(msg);
     this.error(msg);
-    process.exit(1);
+    throw(msg);
```
hayesgm, line 78: Not sure why we changed this (or why we had it). I had issues with the tests not exiting when a node failed, but maybe this is right now?

integration/util/scenario/token.js
```diff
@@ -334,7 +339,7 @@ async function buildTokens(tokensInfoHash, scenInfo, ctx) {
 
   let etherToken = new EtherToken(scenInfo.eth_liquidity_factor, ctx);
   tokens.push(etherToken);
-  if (scenInfo.eth_supply_cap) {
+  if (!!scenInfo.eth_supply_cap) {
```
hayesgm: I think this is identical to the original, so probably can remove this.

integration/util/scenario/validator.js
```diff
@@ -141,63 +162,80 @@ class Validator {
...
+
+  log(text) {
+    if (this.ctx.__deepColor()) {
+      this.ctx.log(this.colorize(`[${this.name}] ${text}`));
+    } else {
+      this.ctx.log(`[${this.colorName()}] ${text}`);
+    }
+  }
+
+  // !!!
```
hayesgm, line 177: Not sure this comment.

integration/util/scenario/validator.js
```diff
@@ -141,63 +162,80 @@ class Validator {
...
-      executionArgs = ['--execution', 'Native'];
-    } else {
-      // Scen did not specify execution. Only force WASM if we're on a non-current version
-      if (!this.version.isCurr()) {
-        executionArgs = ['--execution', 'Wasm'];
-      }
-    }
+    let executionArgs = [
+      '--execution', this.native ? 'Native' : 'Wasm',
+      /*'--wasm-runtime-overrides', this.version.wasmDir()*/
```
hayesgm, line 206: We might want to add this back in? Maybe as an option, at the least?

pallets/cash/src/chains.rs
```diff
@@ -561,9 +550,24 @@ impl ChainBlockTally {
         has_super_majority(&self.dissent, validator_set)
     }
 
-    pub fn has_signer(&self, validator_id: &ValidatorIdentity) -> bool {
+    pub fn has_supporter(&self, validator_id: &ValidatorIdentity) -> bool {
         // note that these set types are not optimized and inefficient
-        self.support.contains(&validator_id) || self.dissent.contains(&validator_id)
+        log!(
```
hayesgm: We can probably remove this. It was for debugging.

pallets/cash/src/internal/events.rs
```diff
@@ -97,6 +97,7 @@ pub fn track_chain_events<T: Config>() -> Result<(), Reason> {
 
 /// Perform the next step of tracking events from an underlying chain.
 pub fn track_chain_events_on<T: Config>(chain_id: ChainId) -> Result<(), Reason> {
+    log!("track_chain_events_on'ing");
```
hayesgm: can probably remove this log, as well.

pallets/cash/src/internal/initialize.rs
```diff
@@ -115,7 +116,7 @@ pub fn initialize_block<T: Config>(now: Timestamp) -> Result<(), Reason> {
     }
 
     <Module<T>>::deposit_event(Event::TransferCash(
-        ChainAccount::Reserved,
+        GATEWAY_VOID,
```
hayesgm, line 119: Not sure what this is?

runtime/src/lib.rs
hayesgm: I think this is right.

scripts/build_release.sh
```diff
@@ -3,6 +3,7 @@
 set -eo pipefail
 
 version="$1"
+machine=`node -e 'console.log(process.platform + "-" + process.arch)'`
```
hayesgm, line 6: That'll do it.

hayesgm (approved):
Awesome stuff.

pallets/cash/src/internal/initialize.rs
jflatow: This is where we transfer from to miners...and now we have `ChainAccount::Gate(...)` instead of `ChainAccount::Reserved`, so its just shorthand for `ChainAccount::Gate([0u8; 32])`

integration/util/scenario/ctx.js
jflatow: Was it doing `throw` before, or you think `throw` is actually broken? We were logging like 3x so I replaced 1 of them and the process exit with a `throw` since it seemed cleaner

PR #355, opened 2021-06-08
make starport more generic

waynenilsen:
Parameterize ETH and ETH: in the starport as immutable

### DIFF
--- a/ethereum/contracts/Starport.sol
+++ b/ethereum/contracts/Starport.sol
@@ -14,8 +14,8 @@ contract Starport {
 
     address immutable public admin;
     address constant public ETH_ADDRESS = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
-    bytes4 constant MAGIC_HEADER = "ETH:";
-    string constant ETH_CHAIN = "ETH";
+    bytes4 immutable chainIdHeader; // = "ETH:" on Ethereum
+    bytes32 immutable chainId; // = "ETH" on Ethereum
     address[] public authorities;
     mapping(address => uint) public supplyCaps;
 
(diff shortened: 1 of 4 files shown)


ethereum/contracts/Starport.sol
```diff
@@ -13,9 +13,9 @@ contract Starport {
     ICash immutable public cash;
 
     address immutable public admin;
-    address constant public ETH_ADDRESS = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
-    bytes4 constant MAGIC_HEADER = "ETH:";
-    string constant ETH_CHAIN = "ETH";
+    address constant public NATIVE_TOKEN_ADDRESS = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
```
hayesgm: This maybe should be immutable and different for different chains, e.g. `0xAaAAaaaaaAAAaaaAAaaaaaaAaAaaaAa` for mAtic?

ethereum/contracts/Starport.sol
```diff
@@ -87,8 +89,8 @@ contract Starport {
      * @dev Use `lock` to lock CASH or collateral assets.
      */
     function lockEthTo(string memory chain, bytes32 recipient) public payable {
-        require(address(this).balance <= supplyCaps[ETH_ADDRESS], "Supply Cap Exceeded");
-        emit Lock(ETH_ADDRESS, msg.sender, chain, recipient, msg.value);
+        require(address(this).balance <= supplyCaps[NATIVE_TOKEN_ADDRESS], "Supply Cap Exceeded");
+        emit Lock(NATIVE_TOKEN_ADDRESS, msg.sender, chain, recipient, msg.value);
```
jflatow: Not sure its worth logging the `NATIVE_TOKEN_ADDRESS` every time?

jflatow:
Looks mostly great, I think the issue with the test is that the constructor isn't being called with the new args which would set the `chainId`? It's funny that saddle is deploying without enough args, but that's what it looks like is happening

ethereum/contracts/Starport.sol
waynenilsen: i think i'm ok with the native token address being the same across all evm chains - don't really see any reason why they should be different

ethereum/contracts/Starport.sol
waynenilsen: yeah i think this just makes lock events generic

waynenilsen:
> Looks mostly great, I think the issue with the test is that the constructor isn't being called with the new args which would set the chainId? It's funny that saddle is deploying without enough args, but that's what it looks like is happening

The issue was the encoding on the JS side as you pointed out in discord. The issue of the "hey, there aren't enough arguments" doesn't happen because `Starport` is not deployed directly it is deployed via `StarportHarness` which provides the additional CLI arguments

jflatow (approved):
Yay awesome 💪  So no interface change, and can be deployed for different native tokens. What are we planning to do about the governance interface (i.e. `executeProposal`)?

waynenilsen:
pulled these changes into #330

this was a good place to review them though and get some consensus

PR #357, opened 2021-06-10
Fix fetch_eth_blocks and formulate reorg issue

jflatow:
We don't need workers to abstain, since that creates different behaviors depending on where the 'invalid blockchain' is found.

Also, avoid fetching duplicates in formulating the reorg.
Also, really only fetch `slack` blocks, not `slack + 1`.

### DIFF
--- a/integration/__tests__/dissent_and_concordance_scen.js
+++ b/integration/__tests__/dissent_and_concordance_scen.js
@@ -88,24 +88,45 @@ buildScenarios('dissent and concordance', scen_info, [
           events: []
         });
 
-        expect(pendingChainBlocks0[0]).toEqual({
-          block: {
-            eth: {
-              hash: block1.hash,
-              parent_hash: block0.hash,
-              number: 1,
-              events: []
-            }
-          },
-          support: [
-            '0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48',
-            '0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d'
-          ],
-          dissent: [
-            '0x306721211d5404bd9da88e0204360a1a9ab8b87c66c1bc2fcdd37f3c2222cc20',
-            '0x90b5ab205c6974c9ea841be688864633dc9ca8a357843eeacf2314649965fe22'
-          ]
-        });
+        if (pendingChainBlocks0[0].block.eth.hash == block1.hash) {
+          expect(pendingChainBlocks0[0]).toEqual({
+            block: {
+              eth: {
+                hash: block1.hash,
+                parent_hash: block0.hash,
+                number: 1,
+                events: []
+              }
+            },
+            support: [
+              '0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48',
+              '0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d'
+            ],
+            dissent: [
+              '0x306721211d5404bd9da88e0204360a1a9ab8b87c66c1bc2fcdd37f3c2222cc20',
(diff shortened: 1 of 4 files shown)


pallets/cash/src/events.rs
```diff
@@ -71,15 +70,9 @@ fn fetch_eth_blocks(
 ) -> Result<ChainBlocks, EventError> {
     debug!("Fetching Eth Blocks [{}-{}]", from, to);
```
hayesgm: Can we revise this to `[{}-{})` if this is truly a right-open interval?

pallets/cash/src/events.rs
```diff
@@ -119,7 +112,7 @@ mod tests {
         ];
 
         let fetch_from = blocks_to_return[0].number;
-        let fetch_to = blocks_to_return[blocks_to_return.len() - 1].number;
+        let fetch_to = blocks_to_return[blocks_to_return.len() - 1].number + 1;
```
hayesgm, line 115: Or this is not: it's actually a closed interval now?

pallets/cash/src/internal/events.rs
```diff
@@ -259,39 +259,51 @@ pub fn formulate_reorg<T: Config>(
     chain_id: ChainId,
     last_block: &ChainBlock,
     true_block: &ChainBlock,
+    chunk_size: u32,
 ) -> Result<ChainReorg, Reason> {
     let starport = get_starport::<T>(chain_id)?;
     let mut reverse_blocks: Vec<ChainBlock> = vec![];
     let mut drawrof_blocks: Vec<ChainBlock> = vec![];
```
hayesgm: Can we just add a comment here `// drawrof is "forward" backwards, since these are forward blocks stored in reverse order` or something?

pallets/cash/src/internal/events.rs
```diff
@@ -259,39 +259,51 @@ pub fn formulate_reorg<T: Config>(
...
+                number.saturating_sub(chunk_size as u64),
                 number,
                 starport,
             )?
             .blocks()
             .collect_rev(),
             None => Vec::new(),
         };
+
+        // its an error if we can't make progress pulling more forward or reverse blocks
```
hayesgm: "it's"

integration/__tests__/dissent_and_concordance_scen.js
```diff
@@ -88,24 +88,45 @@ buildScenarios('dissent and concordance', scen_info, [
...
-          support: [
-            '0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48',
-            '0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d'
-          ],
-          dissent: [
-            '0x306721211d5404bd9da88e0204360a1a9ab8b87c66c1bc2fcdd37f3c2222cc20',
-            '0x90b5ab205c6974c9ea841be688864633dc9ca8a357843eeacf2314649965fe22'
-          ]
-        });
+        if (pendingChainBlocks0[0].block.eth.hash == block1.hash) {
```
hayesgm, line 91: It's pretty funny to me that this is non-deterministic. At 10,000 feet, does that fact feel right?

hayesgm (approved):
lgtm; though I didn't look too deep into the loop related changes.

pallets/cash/src/events.rs
jflatow: Its half-open, we can change the log

pallets/cash/src/internal/events.rs
jflatow: I added a comment about that its just further down

integration/__tests__/dissent_and_concordance_scen.js
jflatow: I think so? Just depends which worker submits its 'truth' first

PR #358, opened 2021-06-10
Add Scen Repl

hayesgm:
Try `yarn repl -i ./repl/testnet.json` to connect a new node to Ropsten. You can reference in the repl anything available on ctx, such as `(await api.query.cash....).toJSON()`, etc etc. Code is very rough right now to test that it might work and how.


jflatow (commented):
This is looking awesome 🔥

PR #362, opened 2021-06-11
Fix TOB #2

waynenilsen:


jflatow (approved):
💪 The bigger question here I think is, do we really need to use strings everywhere?

PR #363, opened 2021-06-11
Fix TOB issue 3 and 4

waynenilsen:


jflatow:
Nice I think this covers both ToB 3 and 4, but can we pull in the tests they added too? (Appendix C)

waynenilsen:
I want to pull in their tests but they're embedded in a PDF in a git diff with the +s etc tried to copy it out but it got completely mangled. Maybe we can ask them for a copy of the patch or something. I can un-mangle but its a bit of manual nonsense

jflatow (approved):
Nice, thanks

PR #364, opened 2021-06-11
build releases on push to develop branch

waynenilsen:


waynenilsen:
closing because only `write`ers can `tag`

PR #365, opened 2021-06-11
build release binaries in github actions

waynenilsen:
tested on `gateway-alpha` result is here 

https://github.com/compound-finance/gateway-alpha/releases/tag/untagged-e2123392f99b00162321

action build is here

https://github.com/compound-finance/gateway-alpha/actions/runs/929596701

### DIFF
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -1,0 +1,96 @@
+name: Build
+on:
+  push:
+    tags:
+      - 'm*'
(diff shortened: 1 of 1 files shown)


hayesgm:
This is _awesome_! Any word on getting cross-arch working?

waynenilsen:
I wanted to get this working first then i will move on to cross compiling - I think on linux it should be fairly straightforward there seems to be more documentation there

PR #366, opened 2021-06-11
Add Gateway address parsing and fix ToB 17

jflatow:
Only parses hex encoded addresses, we can add SS58 later, but we'll have to roll our own (substrate relies on deps we can't pull in).


trx-request/src/lib.rs
```diff
@@ -68,13 +72,28 @@ fn parse_max_amount<'a>(t: &Token) -> Result<MaxAmount, ParseError<'a>> {
...
+            .map_err(|_| ParseError::InvalidChainAccount(Chain::Gate))?;
+        Ok(address)
+    } else {
+        Err(ParseError::InvalidChainAccount(Chain::Gate))?
+    }
+}
+
 fn parse_eth_address<'a>(account: &'a str) -> Result<[u8; 20], ParseError<'a>> {
-    if &account[0..2] != "0x" {
+    if account.len() < 2 || &account[0..2] != "0x" {
```
hayesgm, line 96: This should use starts_with, right?

hayesgm (approved):
lgtm

trx-request/src/lib.rs
jflatow: This one is a little different bc of the clauses below, we should maybe do a larger refactor but didn't want to do that yet

PR #368, opened 2021-06-14
Hardfork m11 Chain Spec

hayesgm:
This patch adds the new chainspec for the m11 hard-fork. This should have the correct values for the new chain, but we should double-check the values.


jflatow (approved):
We could easily be forgetting something, but I can't think of anything

PR #369, opened 2021-06-15
Get rid of regalloc dumps

jflatow:
Actually made the change upstream, but we should probably keep these other changes.

https://github.com/compound-finance/substrate/commit/9bdbf7c0dafc1880eea8ab5fd485b9bd25b5db80

We'll need to rebuild the release artifacts anyway to include them, so might as well mark them with another commit here.


jflatow:
Just found out these changes are being requested upstream: https://github.com/bytecodealliance/regalloc.rs/pull/123

jflatow:
Changes included in #366

PR #370, opened 2021-06-15
Fixes for new camelCase notices.

torreyatcitty:
The latest version of the Polkadot API switched the notice object keys to use camelCase instead of Pascal case. This fixes the notices we missed in our encoding util.

### DIFF
--- a/integration/util/scenario/types.js
+++ b/integration/util/scenario/types.js
@@ -45,8 +45,8 @@ function encodeNotice(notice) {
       },
       [ethNotice.asset, ethNotice.amount, ethNotice.account]
     );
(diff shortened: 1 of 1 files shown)


integration/util/scenario/types.js
```diff
@@ -108,6 +108,7 @@ function encodeNotice(notice) {
       [ethNotice.new_authorities]
     );
   } else {
+    console.log("raw notice is: ", notice);
```
hayesgm: Console log.

hayesgm (approved):
looks great; thanks. One dangling log.

integration/util/scenario/types.js
torreyatcitty: oopsie, removed the dangling.

PR #371, opened 2021-06-17
Fix PolkadotJS Client

hayesgm:
This patch fixes the PolkadotJS client with the patches from @jflatow by using yarn resolutions to grab the published version of those changes from npm. We should still work to get the fixes merged into the mainstream repo of PolkadotJS.


jflatow (approved):
Awesome, this actually works?! 💪

hayesgm:
Yeah, but you should give it a try. "It works on my machine"

PR #374, opened 2021-06-18
Test multiple chunks in formulate reorg

jflatow:


pallets/cash/src/internal/events.rs
```diff
@@ -84,18 +86,18 @@ pub fn risk_adjusted_value<T: Config>(
...
-    let result = match lock.try_lock() {
-        Ok(_guard) => {
-            // Note: chains could be parallelized
-            track_chain_events_on::<T>(ChainId::Eth)
-        }
-
-        _ => Err(Reason::WorkerBusy),
-    };
-    result
+    if TRACK_CHAIN_EVENTS_LOCK.fetch_add(1, Ordering::SeqCst) == 0 {
```
hayesgm: Curious if this provides _different_ guarantees than the previous lock, regardless of the timeout issue? Does the `StorageLock` use the same underlying mechanism?

hayesgm (approved):
looks good; just wondering if there was another change in regards to unavailable data for re-organization.

coburncoburn (approved):
🧼 this looks much tidier and therefore much more likely to work correctly 👏

pallets/cash/src/internal/events.rs
jflatow: Yeah I guess this only worked for native since the memory isn't shared by WASM threads, reverted to 120s deadline with storage lock

hayesgm (approved):
lgtm

PR #376, opened 2021-06-23
Max/check price timestamp

maxwolff:
Check timestamp isn't too far in the future, thus blocking future prices from being posted. fixed trail of bits audit issue 5. Also seperates timestamp logic into its own crate with freeze time logic, so that oracle and cash both depend on it instead of having a circular dependency


pallets/oracle/src/oracle.rs
```diff
@@ -206,6 +208,8 @@ pub fn get_and_check_parsed_price<T: Config>(
     if let Some(last_updated) = PriceTimes::get(&ticker) {
         if parsed.timestamp <= last_updated {
             Err(OracleError::StalePrice)?;
+        } else if parsed.timestamp - last_updated > MAX_PRICE_FUTURE_MS {
```
waynenilsen: I see the check here `if parsed.timestamp <= last_updated {` that prevents underflow in the `else if` but somehow i still find it unnerving to see a naked subtraction with user input so close. This will cause a panic if it underflows.

waynenilsen: this does the same thing without using the subtraction operator

```

    if let Some(last_updated) = PriceTimes::get(&ticker) {
        let time_from_last_update_to_parsed_price = parsed
            .timestamp
            .checked_sub(last_updated)
            .ok_or(OracleError::StalePrice)?;

        if time_from_last_update_to_parsed_price > MAX_PRICE_FUTURE_MS {
            Err(OracleError::TimestampTooHigh)?;
        }
    }
```

pallets/oracle/src/oracle.rs
```diff
@@ -329,6 +334,43 @@ pub mod tests {
         Ok(())
     }
 
+    #[test]
```
waynenilsen, line 343: would love to see another test for `parsed.timestamp == 0` i think it is a conceivable "default value" edge case on the exchange side

waynenilsen (commented):
Generally speaking, I think the implication of this PR is that if we don't post a price for 100 seconds we can never post a price again, right? I think that is probably something that we also don't really want to do to ourselves.

Since these payloads are signed the only real bad thing that I could see happening is a bug on the exchange side regarding time units, something like "oops we multiplied by 10 too many times".

The change is good but I think we should make `MAX_PRICE_FUTURE_MS` something like 3 months to be conservative? Then we are going to find it pretty difficult to lock ourselves out esp before launch.

Also, once we merge this for testnet we need to set up an automated poster service to keep us from getting locked out

jflatow (changes requested):
Great work, and also agree with the comments from 🦇 and resolution 🐺 mentioned about checking too high timestamp relative to current time (this is what we did on Eth too right?)

timestamp/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "timestamp"
+version = "0.1.0"
+authors = ['Compound <https://compound.finance>']
+edition = "2018"
```
coburncoburn, line 5: I don't know what to tell you this... but 3 years have passed

pallets/cash/src/lib.rs
```diff
@@ -101,8 +99,10 @@ pub trait Config:
         + UnfilteredDispatchable<Origin = Self::Origin>
         + GetDispatchInfo;
 
-    /// Convert implementation for Moment -> Timestamp.
-    type TimeConverter: Convert<<Self as pallet_timestamp::Config>::Moment, Timestamp>;
+    /// Gets the most recent timestamp and converts it from a moment
+    type GetConvertedTimestamp: timestamp::GetConvertedTimestamp<
+        <Self as pallet_timestamp::Config>::Moment,
```
coburncoburn, line 104: kind of meh that cash still has a direct dependency on pallet_timestamp. Is it possible to paper over this part too?

coburncoburn: as in, can Timestamp: The Crate take over completely so that Cash: The Pallet doesn't know about Timestamp: The Pallet

maxwolff: hm maybe

coburncoburn (approved):

pallets/cash/src/lib.rs
jflatow: don't we also use timestamps for calculating yield?

pallets/oracle/src/tests/mod.rs
```diff
@@ -56,6 +56,7 @@ fn test_post_price_happy_path() {
     let test_signature = hex::decode("41a3f89a526dee766049f3699e9e975bfbabda4db677c9f5c41fbcc0730fccb84d08b2208c4ffae0b87bb162e2791cc305ee4e9a1d936f9e6154356154e9a8e9000000000000000000000000000000000000000000000000000000000000001c").unwrap();
     new_test_ext().execute_with(|| {
         initialize_storage(); // sets up ETH
+        <pallet_timestamp::Pallet<Test>>::set_timestamp(500);
```
jflatow, line 59: Should this be part of initialize_storage?

timestamp/Cargo.toml
jflatow: Still thats the rust edition...

coburncoburn:
Right 👷‍♂️

PR #378, opened 2021-06-24
Don't Request All Block Transactions

hayesgm:
This patch updates our Etheruem Client code to _not_ ask for transactions, since we instead ask for logs via a different API. This should heavily reduce the total size of the responses we're getting from Ethereum, since there could be 100's of unrelated transctions in a given block that we now just get the hashes for.

See: https://eth.wiki/json-rpc/API#eth_getblockbynumber


coburncoburn (approved):
definitely good to get less data here

jflatow (approved):
💪 Great improvement

PR #382, opened 2021-06-25
Ugliness necessary to prevent the extra block fetch from worker

jflatow:


coburncoburn (commented):
this is just to prevent the literal fetching of it from infura twice, but doesn't actually change the output of this function, right?

jflatow:
> this is just to prevent the literal fetching of it from infura twice, but doesn't actually change the output of this function, right?

Yeah it doesn't change what the blocks we process in the runtime are, but changes what blocks we fetch and when

pallets/cash/src/internal/events.rs
```diff
@@ -102,29 +102,41 @@ pub fn track_chain_events_on<T: Config>(chain_id: ChainId) -> Result<(), Reason>
...
         let event_queue = get_event_queue::<T>(chain_id)?;
         let slack = queue_slack(&event_queue) as u64;
-        let blocks = fetch_chain_blocks(
-            chain_id,
-            last_block.number() + 1,
-            last_block.number() + 1 + slack,
-            starport,
-        )?
-        .filter_already_supported(&me.substrate_id, pending_blocks);
+        let blocks = next_block.concat(
```
coburncoburn, line 118: feels like we should at least verify the parent_hash chain when fetching here, similar to the way re-orgs do. Am I missing something?

jflatow: The reorgs don't verify, the chain is ultimately checked by the runtime

PR #383, opened 2021-06-25
flexible CLI and ENV based configuration

waynenilsen:
This provides future proof configuration for validators. No longer will additional validator specific configuration upgrades (eg adding node RPC URL) need validators to update their binary, soft fork WITH validator setting new ENV is sufficient.

TL;DR

This is a breaking change

before

```shell
./gateway --chain ... --base-path ... ...... \
   --eth-rpc-url http://localhost:8004 \
   --eth-key-id my_eth_key_id \
   --miner Eth:0x8AD1b2918C34EE5d3E881A57c68574EA9dbEcB81 \ 
   --opf-url http://127.0.0.1:58718/
```

after

```shell
./gateway --chain ... --base-path ... ...... --env \
   ETH_RPC_URL=http://localhost:8004 \
   ETH_KEY_ID=my_eth_key_id \
   MINER=Eth:0x8AD1b2918C34EE5d3E881A57c68574EA9dbEcB81 \ 
   OPF_URL=http://127.0.0.1:58718/
```

The big difference is that we can now add whatever CLI args we want and read them via the runtime without needing to upgrade the node binary.

### DIFF
--- a/integration/util/scenario/validator.js
+++ b/integration/util/scenario/validator.js
@@ -220,12 +220,15 @@ class Validator {
     let newCliArgs = [];
     let ethRpcUrl = this.ethProxy ? this.ethProxy.serverUrl() : this.ctx.eth.web3Url;
     if (this.version.supports('full-cli-args')) {
-      newCliArgs = [
+      if (this.version.supports('generic-cli-args')) {
+        newCliArgs.push('--')
+      }
+      newCliArgs.push(...[
         '--eth-rpc-url', ethRpcUrl,
         '--eth-key-id', "my_eth_key_id",
         '--miner', `Eth:${this.ethAccount}`,
         '--opf-url', this.ctx.__opfUrl(),
-      ];
+      ]);
     } else {
       env['ETH_RPC_URL'] = ethRpcUrl;
       env['ETH_KEY_ID'] = "my_eth_key_id";
@@ -249,13 +252,13 @@ class Validator {
       '--no-mdns',
       '--node-key',
       this.nodeKey,
-      ...newCliArgs,
       '-laura=info,executor=info,runtime=info,gateway=info,pallet_cash=info,session=info',
       '--reserved-only',
       ...executionArgs,
       ...this.bootnodes,
       ...this.extraArgs,
-      ...this.validatorArgs
+      ...this.validatorArgs,
+      ...newCliArgs,
     ], { env });
 
     process.on('exit', () => {
--- a/integration/util/scenario/validator.js
+++ b/integration/util/scenario/validator.js
@@ -221,14 +221,21 @@ class Validator {
     let ethRpcUrl = this.ethProxy ? this.ethProxy.serverUrl() : this.ctx.eth.web3Url;
     if (this.version.supports('full-cli-args')) {
       if (this.version.supports('generic-cli-args')) {
-        newCliArgs.push('--')
+        newCliArgs = [
+          '--gateway-args',
+          'eth-rpc-url', ethRpcUrl,
+          'eth-key-id', "my_eth_key_id",
+          'miner', `Eth:${this.ethAccount}`,
+          'opf-url', this.ctx.__opfUrl(),
+        ];
+      } else {
+        newCliArgs = [
+          '--eth-rpc-url', ethRpcUrl,
+          '--eth-key-id', "my_eth_key_id",
+          '--miner', `Eth:${this.ethAccount}`,
+          '--opf-url', this.ctx.__opfUrl(),
+        ];
       }
-      newCliArgs.push(...[
-        '--eth-rpc-url', ethRpcUrl,
-        '--eth-key-id', "my_eth_key_id",
-        '--miner', `Eth:${this.ethAccount}`,
-        '--opf-url', this.ctx.__opfUrl(),
-      ]);
     } else {
       env['ETH_RPC_URL'] = ethRpcUrl;
       env['ETH_KEY_ID'] = "my_eth_key_id";
--- a/integration/util/scenario/validator.js
+++ b/integration/util/scenario/validator.js
@@ -222,11 +222,11 @@ class Validator {
     if (this.version.supports('full-cli-args')) {
       if (this.version.supports('generic-cli-args')) {
         newCliArgs = [
-          '--gateway-args',
-          'eth-rpc-url', ethRpcUrl,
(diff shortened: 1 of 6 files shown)


integration/util/scenario/validator.js
```diff
@@ -218,12 +218,15 @@ class Validator {
     let newCliArgs = [];
     let ethRpcUrl = this.ethProxy ? this.ethProxy.serverUrl() : this.ctx.eth.web3Url;
     if (this.version.supports('full-cli-args')) {
-      newCliArgs = [
+      if (this.version.supports('generic-cli-args')) {
+        newCliArgs.push('--')
```
waynenilsen: really wish i could do this without needing this small breaking change :(

pallets/runtime-interfaces/src/lib.rs
```diff
@@ -48,6 +51,27 @@ const MINER_DEFAULT: &str = "Eth:0x0000000000000000000000000000000000000000";
 const ETH_RPC_URL_DEFAULT: &str = "https://ropsten-eth.compound.finance";
 const OPF_URL_DEFAULT: &str = "https://prices.compound.finance/coinbase";
 
+fn validator_config_interface_get_internal(key: &str) -> Option<String> {
+    // check env override
+    if let Ok(eth_key_id_from_env) = std::env::var(key) {
```
coburncoburn: wouldn't be eth_key_id necessarily here right? just whatever was requested from the key?

pallets/runtime-interfaces/src/lib.rs
```diff
@@ -60,79 +84,32 @@ const OPF_URL_DEFAULT: &str = "https://prices.compound.finance/coinbase";
...
-        }
-        // check config
-        if let Ok(config) = VALIDATOR_CONFIG.lock() {
-            if let Some(inner) = config.as_ref() {
-                return Some(inner.miner.clone().into());
-            }
-        }
-        // not set
-        return Some(MINER_DEFAULT.into());
+        validator_config_interface_get_internal(MINER_ENV_VAR).map(Into::into)
```
coburncoburn, line 112: definitely tidies up these readers!

integration/util/scenario/versions.js
```diff
@@ -183,6 +183,7 @@ class Version {
       'full-cli-args': (v) => v >= 9,
       'eth-starport-parent-block': (v) => v >= 9,
       'new-artifacts': (v) => v >= 10,
+      'generic-cli-args': (v) => v >= 12, // todo: not sure exactly what this should be
```
coburncoburn, line 186: yeah depends on when it is merged I guess...

integration/util/scenario/validator.js
jflatow: Why do we need to support multiple cli versions? Just for upgrade tests?

waynenilsen: it is mostly unfortunate that we will need to tell any validators that are using the cli args to move these args to the end after a `--` this code change is to support the upgrade tests

pallets/runtime-interfaces/src/lib.rs
waynenilsen: yeah good catch, this just came out of the refactor

node/src/cli.rs
```diff
@@ -1,23 +1,46 @@
...
+    // collect args after -- like cargo, eg cargo run -- --help for generic key-value configuration
+    // expected format is --eth-key-id abcdefg --miner 123456789
+    // OR
+    // ETH_KEY_ID abcdefg MINER 123456789
+    // equals = linking not supported as in --eth-key-id=abcdefg
+    #[structopt()]
+    pub args: Vec<String>,
+}
+
+impl GatewayCmd {
```
jflatow, line 19: Hmm can we just use a StructOpt with all of the ones we accept, and then convert that into a dynamic HashMap, so we can see what options are currently supported and avoid this parsing logic?

waynenilsen: The whole issue of `parse` is so that we can have env vars the way we expect them to be and CLI args the way we expect them to be - eg `--eth-rpc-url` on the cli and `ETH_RPC_URL` in the environment. This allows us to do overrides in a very consistent way while providing some level of backward compatibility.

We could use structopt for the ones we already accept but i thought it would be more confusing to have to add _only the new CLI params added after this pr_ after `--` the whole issue of the `--` thing is confusing enough on its own without having to say `--eth-rpc-url abcdef -- --flow-rpc-url ghijkl` its really easy to mess that up.

pallets/runtime-interfaces/src/lib.rs
waynenilsen: thanks 03096ee

node/src/cli.rs
jflatow: I guess I'm confused...what are the old ones?

waynenilsen: There are currently 4 cli params that we add for gateway, those, in my scenario, are the "old ones". What i'm saying is that if we keep those 4 the way they are then when we require a "new" one, eg `--flow-rpc-url` we will need to put it _after_ the `--` so it will look like this 

```
#                    vvvvv existing params vvvvvv          vvvv all "new" params would have to go after the --
./gateway ... --eth-rpc-url ...                            --   --flow-rpc-url ... 
```

the way that I implemented it is 

```
#               vvvvv substrate params vvvvv           vvvv   all gateway params vvvvv
./gateway ... --unsafe-rpc ...                       --      --eth-rpc-url http://...  --flow-rpc-url ... 
```

hopefully this illustration clears up this question for you

_that_ is why I made _all_ the params "dynamic"

jflatow: It makes sense to me that all our custom Gateway params come after the `--`, but how does that break using a StructOpt to describe what they are (and converting that to a HashMap)?

waynenilsen: The short answer to your question is that the two are mutually exclusive. The literal function of the `--` is "throw the rest into the last `Vec<String>` bag". Hopefully this makes sense?

jflatow: I guess I was thinking we could just take the throw away bag and use StructOpt to parse that separately, but maybe thats now how it works. Anyway sounds like `-e` will be the preferred method anyway and docs have list of accepted params (although I am sure those will get out of date easily)

jflatow (approved):
Very nice 🧼

PR #390, opened 2021-06-30
add support for gateway addresses in rpcs

coburncoburn:


jflatow (approved):

gateway-crypto/src/no_std.rs
```diff
@@ -199,6 +199,17 @@ pub fn eth_hash_string(eth_hash: &[u8; 32]) -> String {
     format!("0x{}", hex::encode(eth_hash))
 }
 
+pub fn gateway_str_to_address(address_str: &str) -> Option<[u8; 32]> {
+    if address_str.len() == 66 && &address_str[0..2] == "0x" {
+        if let Ok(bytes) = hex::decode(&address_str[2..66]) {
```
jflatow, line 204: I guess we'll fix up / fuzz up all these string things in a future pass

coburncoburn: yeah I think one function "hex from str" would be a better way

PR #393, opened 2021-07-02
Add Stablenet

hayesgm:
This patch adds support for Stablenet connected to Goerli. We will start with a single node: "Sally" that will run the net and we are _not_ using Governance for the time-being. Georli shall reign supreme.


Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN chmod +x releases/m16/gateway-linux-x86 
+
+CMD releases/m16/gateway-linux-x86 \
+  --chain chains/stablenet/chain-spec-raw.json \
+  --base-path /chain \
+  --ws-external \
+  --rpc-methods Unsafe \
+  --rpc-external \
+  --no-mdns \
+  --log runtime=trace,pallet_cash=trace,ethereum_client=debug \
```
jflatow, line 22: Default is trace? I feel like we should only add tracing if we are looking to debug something in particular

scripts/pull_release.sh
```diff
@@ -21,7 +21,7 @@ release_dir="./releases/$version"
 
 mkdir -p "$release_dir"
 
-files=("gateway_runtime.compact.wasm" "gateway_runtime.checksum" "types.json" "rpc.json" "contracts.json")
+files=("contracts.json" "gateway-darwin-arm64" "gateway-darwin-arm64.checksum" "gateway-linux-x86" "gateway.wasm" "gateway.wasm.checksum" "rpc.json" "types.json")
```
jflatow, line 24: Pull darwin but not linux? I guess this is meant only for running on mac laptops?

jflatow (approved):

chains/stablenet/chain-config.json
```diff
@@ -0,0 +1,74 @@
+{
+  "eth_network": "goerli",
+  "base_chain": "testnet",
+  "websocket": "wss://sarge.compound.cash",
+  "rpc": "http://sarge-rpc.compound.cash:5000",
+  "validators": [
+    {
+      "substrate_id": "5H6Wzjt1nQ5aFx5E4FDVnpYDCWb1JHcQSwHgdqUyiiZUdbc9",
```
coburncoburn, line 8: where is this id from ?

jflatow (approved):

PR #394, opened 2021-07-06
Fix the defaults in CLI

waynenilsen:
* during the CLI refactor a bug introduced broke default values for CLI parameters
* this PR fixes that and adds a unit test for the interesting cases


pallets/runtime-interfaces/src/lib.rs
```diff
@@ -63,7 +62,9 @@ fn validator_config_interface_get_internal(key: &str) -> Option<String> {
     // check config
     if let Ok(config) = VALIDATOR_CONFIG.lock() {
         if let Some(inner) = config.as_ref() {
-            return inner.map.get(key).map(Clone::clone);
+            if let Some(x) = inner.map.get(key) {
+                return Some(x.clone());
+            }
```
jflatow, line 67: Sneaky

jflatow (approved):

PR #395, opened 2021-07-06
[COMP-1278] Limit trx request size

waynenilsen:
Set a cap on the size of these in validate_unsigned to prevent people DoSing validators

### DIFF
--- a/pallets/cash/src/params.rs
+++ b/pallets/cash/src/params.rs
@@ -51,3 +51,6 @@ pub const ERROR_WEIGHT: u64 = 100_000_000;
 
 /// The void account from whence miner CASH is transferred out of.
 pub const GATEWAY_VOID: ChainAccount = ChainAccount::Gate([0u8; 32]);
+
+/// The maximum length of a trx request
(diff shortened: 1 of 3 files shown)


pallets/cash/src/params.rs
```diff
@@ -51,3 +51,6 @@ pub const ERROR_WEIGHT: u64 = 100_000_000;
 
 /// The void account from whence miner CASH is transferred out of.
 pub const GATEWAY_VOID: ChainAccount = ChainAccount::Gate([0u8; 32]);
+
+/// The maximum length of a trx request
+pub const MAX_TRX_REQUEST_LEN: usize = 2048; // xxx todo: what length do we want here?
```
jflatow: We could go with this, or even cut it in half I think

coburncoburn (approved):
small & straightforward 👍

jflatow (commented):
Very nice 💪 , let's just agree about the size before merging and remove the XXX - personally I am ✅  with the limit you have

maxwolff (approved):

PR #396, opened 2021-07-06
[COMP-1277] Replace miner env var with 'current validator' gateway address

jflatow:


jflatow:
Not at all clear how we're supposed to get a `<T: Config>` to an `InherentDataProvider`, and overall quite ugly 😞

node/src/service.rs
```diff
@@ -100,6 +100,7 @@ pub fn new_partial(
             justification_import: Some(Box::new(grandpa_block_import.clone())),
             client: client.clone(),
             create_inherent_data_providers: move |_, ()| async move {
+                // XXX how can we pass a <T: Config>??
                 let miner = pallet_cash::internal::miner::InherentDataProvider {};
```
apopiak, line 104: in principal you can depend on the `gateway-runtime::Runtime` struct and pass it to the generic here, that binds your node impl to a particular (set of) runtime version(s) though

pallets/cash/src/internal/miner.rs
```diff
@@ -52,24 +55,16 @@ impl std::ops::Deref for InherentDataProvider {
...
-
-        let miner_address = our_std::str::from_utf8(&miner_address_str).map_err(|_| {
-            sp_inherents::Error::Application(Box::from("invalid miner address bytes"))
-        })?;
-
-        let chain_account: ChainAccount = our_std::str::FromStr::from_str(miner_address)
-            .map_err(|_| sp_inherents::Error::Application(Box::from("invalid miner address")))?;
-
-        inherent_data.put_data(INHERENT_IDENTIFIER, &chain_account)
+        let validator = core::get_current_validator::<T>()
```
apopiak, line 65: this is a function accessing runtime storage I assume?
I think you'll need to introduce a runtime api that allows you to access that data (and that doesn't depend on a `<T: Config>`) similar to e.g. the `GrandpaApi` here: https://github.com/paritytech/substrate/blob/4e75f511e1357cf54c31d9e82d0cf3e39a2b8c28/bin/node/runtime/src/lib.rs#L1307

PR #398, opened 2021-07-07
[COMP-1282] Filter Transfer 0 events emitted when miners are paid

waynenilsen:
These are just cluttering up the subscan with noise (and will do the same for other tools)

### DIFF
--- a/pallets/cash/src/internal/initialize.rs
+++ b/pallets/cash/src/internal/initialize.rs
@@ -115,12 +115,16 @@ pub fn initialize_block<T: Config>(now: Timestamp) -> Result<(), Reason> {
         }
     }
 
-    <Module<T>>::deposit_event(Event::TransferCash(
-        GATEWAY_VOID,
-        last_miner,
-        last_miner_share_principal,
-        cash_index_new,
-    ));
+    if last_miner_share_principal != CashPrincipalAmount::ZERO {
+        // No need to emit events when nothing happens
+        <Module<T>>::deposit_event(Event::TransferCash(
+            GATEWAY_VOID,
+            last_miner,
+            last_miner_share_principal,
+            cash_index_new,
+        ));
+    }
+
     <Module<T>>::deposit_event(Event::MinerPaid(last_miner, last_miner_share_principal));
 
     Ok(())
@@ -164,6 +168,7 @@ mod tests {
 
             let result = initialize_block::<Test>(now);
             let shares = CashPrincipalAmount(242097062);
+            let cash_index = CashIndex::from_nominal("1.192441828000000000");
             assert_eq!(result, Ok(()));
 
             assert_eq!(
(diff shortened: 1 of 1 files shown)


waynenilsen:
substrate coverage failure above looks like a flaky test - it runs on my machine and the line is the assertion below

```
            // Third round should process
            let pending_blocks = PendingChainBlocks::get(ChainId::Eth);
            assert_eq!(pending_blocks.len(), 0);
```

for context

jflatow:
> substrate coverage failure above looks like a flaky test - it runs on my machine and the line is the assertion below
> 
> ```
>             // Third round should process
>             let pending_blocks = PendingChainBlocks::get(ChainId::Eth);
>             assert_eq!(pending_blocks.len(), 0);
> ```
> 
> for context

AFAIK this test shouldn't be flaky

pallets/cash/src/internal/initialize.rs
```diff
@@ -115,12 +115,17 @@ pub fn initialize_block<T: Config>(now: Timestamp) -> Result<(), Reason> {
...
+        // These are just cluttering up tools with noise
+        <Module<T>>::deposit_event(Event::TransferCash(
+            GATEWAY_VOID,
+            last_miner,
+            last_miner_share_principal,
+            cash_index_new,
+        ));
+    }
+
     <Module<T>>::deposit_event(Event::MinerPaid(last_miner, last_miner_share_principal));
```
jflatow, line 128: Should we filter this one too?

pallets/cash/src/internal/initialize.rs
```diff
@@ -115,12 +115,17 @@ pub fn initialize_block<T: Config>(now: Timestamp) -> Result<(), Reason> {
...
     }
 
-    <Module<T>>::deposit_event(Event::TransferCash(
-        GATEWAY_VOID,
-        last_miner,
-        last_miner_share_principal,
-        cash_index_new,
-    ));
+    if last_miner_share_principal != CashPrincipalAmount::ZERO {
+        // Filter Transfer 0 events emitted when miners are paid
```
jflatow: Maybe just "No need to emit events when nothing happens", these wont be cluttering anything up once this is merged

pallets/cash/src/internal/initialize.rs
waynenilsen: Yeah i wasn't sure on this one that is not what was described in the ticket. I think it probably makes sense to have one of these per block, something like our version of a coinbase tx event

waynenilsen:
hmm now `substrate coverage` went ✅ but `ethereum tests` went from ✅  to ❌ 

```
FAIL tests/cash_token_test.js (8.269s)
  ● CashToken › Upgradeable › should be upgradeable to new logic (without call)

    expected (0x6f564ea7efdd90efad8923c6a706428e2cc1e3) == (0x006F564eA7eFDd90eFAD8923C6a706428E2cC1e3)

      87 |       await proxyAdmin.methods.upgrade(cash._address, cashImpl2._address).send({ from: root });
      88 | 
    > 89 |       expect(await getProxyImplementation(cash)).toMatchAddress(cashImpl2._address);
         |                                                  ^
      90 |       expect(await getProxyAdmin(cash)).toMatchAddress(proxyAdmin._address);
      91 | 
      92 |       expect(await cash.methods.admin().call()).toEqual(account3);

      at Object.<anonymous> (tests/cash_token_test.js:89:50)
          at runMicrotasks (<anonymous>)
```

🤷 will have to look into this more .....

jflatow (approved):

PR #399, opened 2021-07-07
WIP: Palletize Validators

hayesgm:


jflatow:
Thinking we should just kill this one for now since we are working on refactors elsewhere?

PR #401, opened 2021-07-13
Use build_release.sh script

mykelp:
This patch updates the `release-workflow` to utilize the `build_release.sh` script, and uploads all the necessary files to the newly created release.

### DIFF
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -36,7 +36,7 @@ jobs:
         run: |
           set -eo pipefail
           ./scripts/get_substrate.sh --fast
-          cargo build --release
+          ./scripts/build_release.sh mRelease
 
       - name: Upload Release Asset
         id: upload-gateway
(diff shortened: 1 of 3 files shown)


jflatow (approved):
Ok very nice, I think this does everything we want?

1. We kick it off by pushing the `mXXX` tag (presumably after a version bump commit)
2. This creates a draft PR, we can go and edit in the meantime
3. An hour or so later the artifacts are done uploading to the PR, we can publish before or after this is done

hayesgm (approved):
lgtm. only question is if we should just use like `m9999` instead of `mRelease` so if one day some script decides to parse the number it doesn't barf?

jflatow:
> lgtm. only question is if we should just use like `m9999` instead of `mRelease` so if one day some script decides to parse the number it doesn't barf?

Equally bad and arguably worse to use a number unless its guaranteed to be a safe number? Maybe we can just designate in the parser that it should be compatible with this (and easy to enforce with this now going in CI)?

mykelp:
@jflatow This workflow will create a draft/prerelease not a PR. Did we need it to create a PR?

hayesgm:
> > lgtm. only question is if we should just use like `m9999` instead of `mRelease` so if one day some script decides to parse the number it doesn't barf?
> 
> Equally bad and arguably worse to use a number unless its guaranteed to be a safe number? Maybe we can just designate in the parser that it should be compatible with this (and easy to enforce with this now going in CI)?

FWIW, we actually already use 9999 to indicate "current" in scenarios, so it's pretty safe to say if we get there we'll need to rethink that number. This practice actually comes from terraform, as well.

PR #402, opened 2021-07-14
Add Stack Trace from Call-site not Error site

hayesgm:
Due to the async callback nature of sendAndWaitForEvents, errors are raised in the callback code, which leads to an error's stacktrace being just that, in some generic callback code. This patch captures a backtrace at the original callsite and then attaches that to an error before it's returned. As such, it's possible to give the user a more helpful error. It's a bit of a hack, but not utilizing any undefined behavior.

### DIFF
--- a/integration/util/scenario/event_tracker.js
+++ b/integration/util/scenario/event_tracker.js
@@ -119,6 +119,7 @@ class EventTracker {
   }
 
   sendAndWaitForEvents(call, opts = {}) {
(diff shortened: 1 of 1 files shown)


jflatow (approved):
Makes sense 🔥 Better than nothing assuming its accurate, and don't see why it wouldn't be

Khaiphan11022 (approved):

coburncoburn (approved):

PR #404, opened 2021-07-15
Update test workflow to not run audit and integrations

mykelp:
The goal of this patch was to make progress towards following a correct CI workflow. Currently our cargo audit job always fails based on dependencies for Substrate that we don't control. Our integration tests take well over an hour to give success/failure results, leading to PRs being merged before waiting for them to pass. This patch adjusts the `test-workflow` for GitHub actions to skip running the `cargo audit` and `integration` jobs. It also fixes a bug in `ethereum-coverage` that should now return proper test/pass results. We can hopefully begin to follow a pattern of only merging PRs with ✅, while we devise a strategy to improve our integrations testing for inclusion in CI.

### DIFF
--- a/ethereum/package.json
+++ b/ethereum/package.json
@@ -11,7 +11,7 @@
     "postinstall": "cd node_modules/solidity-parser-antlr && yarn install && yarn build",
     "compile": "npx saddle compile",
     "test": "npx saddle compile && npx saddle test",
-    "coverage": "npx saddle compile --trace && npx saddle coverage && npx istanbul report --root=./coverage lcov json",
+    "coverage": "BUILD_FILE=.build/contracts-trace.json npx saddle compile --trace && npx saddle coverage && npx istanbul report --root=./coverage lcov json",
     "console": "NODE_OPTIONS='--experimental-repl-await' npx saddle console",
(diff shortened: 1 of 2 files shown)


ethereum/package.json
```diff
@@ -11,7 +11,7 @@
     "postinstall": "cd node_modules/solidity-parser-antlr && yarn install && yarn build",
     "compile": "npx saddle compile",
     "test": "npx saddle compile && npx saddle test",
-    "coverage": "npx saddle compile --trace && npx saddle coverage && npx istanbul report --root=./coverage lcov json",
+    "coverage": "BUILD_FILE=.build/contracts-trace.json npx saddle compile --trace && BUILD_FILE=.build/contracts-trace.json npx saddle coverage && npx istanbul report --root=./coverage lcov json",
```
hayesgm, line 14: FYI, you probably don't need the second `BUILD_FILE` env var here.

hayesgm (approved):

jflatow (approved):
💪 Amazing thanks 🥬 !

ethereum/package.json
mykelp: No, unfortunately it needs it

